Issue #309, opened 2019-07-11
Error - File could not be opened


chraibi:
In Gitlab by @anna-braun on Apr 17, 2019, 11:11 [[origin](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/issues/316)]

@chraibi 
We talked about this problem and I have tried a lots of variation, all with the same result, the simulation fails because a file (writing door statistics) can not be opened.
![Bildschirmfoto_2019-04-16_um_16.31.50](https://gitlab.version.fz-juelich.de/JuPedSim/jpscore/uploads/cc9c4cc4139b4c863b5be1711864caf0/Bildschirmfoto_2019-04-16_um_16.31.50.png)

Changing the size of the files (door-statistics) for 10 MB to 1 MB, had no effect. Also limiting the number of people to a minimum.
As I did simulations without sources, which failed as well, we can exclude them as "source of error".
Also normal goals can be excluded. Simulation with WAs failed when using more than one WA. So maybe there is a problem?

[correct_Bahnsteige.xml](/uploads/b45db6b2baa8b324ccbcea3dfc9ac219/correct_Bahnsteige.xml)

[Personen_ini.xml](/uploads/8ffa94f6ba0b88008821814fdd72b3e5/Personen_ini.xml)

[log.txt](/uploads/4a43f79e41068712a55fbecfc46040de/log.txt)

chraibi:
In Gitlab by @anna-braun on Apr 17, 2019, 11:13

changed the description

chraibi:
In Gitlab by @chraibi on Apr 29, 2019, 11:40

assigned to @chraibi

chraibi:
In Gitlab by @chraibi on May 4, 2019, 13:51

Another "source of error" could be simply lack of memory. 

I tried to reproduce this error with 1MB limit *and* `statistics=true`, but without any luck. 

![Screenshot_2019-05-04_at_13.49.07](https://gitlab.version.fz-juelich.de/JuPedSim/jpscore/uploads/2c4478466cf5f8ac322dcd10db83b1c5/Screenshot_2019-05-04_at_13.49.07.png)

Can we have a look at the output of `df -h`?

![Screenshot_2019-05-04_at_13.49.54](/uploads/f3b1f97b95b52b8b4128e5aa8842b309/Screenshot_2019-05-04_at_13.49.54.png)

chraibi:
In Gitlab by @anna-braun on May 6, 2019, 08:25

[...]

PR #417, opened 2019-07-17
Fix memory/resource leak in Simulation class

Ozaq:
PrintStatistics was leaking FileHandler objects by not deleting them.
Instead of creating them on the heap they are now created on the stack
and autmatically removed at the end of the enclosing scope.

Partially addresses issue: 309

See my comment in #309


chraibi:
@Ozaq  very nice! hanks. This bug was so annoying...

Ozaq:
Well this does not fix it all, the same bug exists still with writing out transitions. I’ll see that I can get a Pr out for that soon

Issue #414, opened 2019-07-17
Warnings and code style in WaitingAreas


chraibi:
It would be good to fix these warnings:

<img width="1225" alt="Screenshot 2019-07-17 11 47 59" src="https://user-images.githubusercontent.com/5772973/61345732-dd1a0780-a888-11e9-94bb-ed5a7942fb5a.png">

Besides, in the throughout the code, the convention of appending a `_` to member variables is used.  Therefore, it would be also good to stick to this convention in this [class](https://github.com/JuPedSim/jpscore/blob/develop/geometry/WaitingArea.h)

schroedtert:
Closed with 7d5ed737646c0313760eb64de80f6b20fa5ace25

Issue #413, opened 2019-07-17
Unnecessary dependency on Boost/random


chraibi:
Any special reason why in [Trips](https://github.com/JuPedSim/jpscore/blob/develop/geometry/Trips.cpp#L6) Boost is used to produce random numbers?

I think this functionality is standard in [C++11](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine) and should be used instead.

schroedtert:
Closed with 7d5ed737646c0313760eb64de80f6b20fa5ace25

Issue #382, opened 2019-07-11
Writing Errors for WAs


chraibi:
In Gitlab by @anna-braun on Jun 24, 2019, 13:03 [[origin](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/issues/330)]

In issue #327 we had to spend a lot of time to find our problem and in the end it was just one wrong letter in the definition of a waiting area. 
As we have mentioned in #327 the simulation failed, without writing an error to the log-file. In the huge files for KapaKrit we might have 3.000 lines or more. Finding a typing error without any message in the log-file could take hours.   
So, @schroedtert, is it possible to write such errors into the log? It does not have to be a very specified log-message, which tells you the specific line or something. But a small hint to see in which goal/waiting area the error occurs would be very helpful. Maybe by writing the goals id to the log.

chraibi:
any updates here?

schroedtert:
Closed with 7d5ed737646c0313760eb64de80f6b20fa5ace25

PR #419, opened 2019-07-18
Error messages, documentation and clean up waiting areas

schroedtert:
- Added improved error messages when parsing waiting areas (Solves #382)
- Added documentation to WaitingArea, removed warnings and follow code convention (Solves #414)
- Clean up of some unused classes (Trips) (Solves #413)


geometry/GoalManager.cpp
```diff
@@ -14,57 +36,36 @@ void GoalManager::SetGoals(std::map<int, Goal*> goals)
...
-//                    //        update for goal where next goal has goal id
-//               }
-//          }
-//     }
-
      // Ped is in current waiting area
      if (CheckInsideWaitingArea(ped, ped->GetFinalDestination())){
           WaitingArea* wa = dynamic_cast<WaitingArea*>(_allGoals[ped->GetFinalDestination()]);
-          wa->addPed(ped->GetID());
+          // Add ped to waiting area
```
Ozaq: This comment is not really necessary, it is just repeating the code below.

geometry/GoalManager.cpp
```diff
@@ -14,57 +36,36 @@ void GoalManager::SetGoals(std::map<int, Goal*> goals)
...
+          wa->AddPed(ped->GetID());
+          // Ped enters waiting area
```
Ozaq: See above

geometry/GoalManager.cpp
```diff
@@ -14,57 +36,36 @@ void GoalManager::SetGoals(std::map<int, Goal*> goals)
...
           ped->EnterGoal();
-          if (!wa->isOpen()) {
+
+          // Check state of waiting area
+          if (!wa->IsOpen()) {
                SetState(wa->GetId(), false);
           }
      }
 
+     // Ped already had a goal AND has not reached final goal
```
Ozaq: Instead of this comment you could rewrite the code to something more readable like:
```
const bool pedHasGoal = ped->GetLastGoalID() != -1 ;
const bool pedHasNotReachedFinalGoal = ped->GetFinalDestination() != ped->GetLastGoalID());
if(pedHasGoal && pedHasNotReachedFinalGoal) {
    // ... 
}

File: IO/GeoFileParser.cpp [hunk 4]
```diff
@@ -733,37 +757,39 @@ Goal* GeoFileParser::parseWaitingAreaNode(TiXmlElement * e)
      Log->Write("INFO:\t  Goal subroom_id: %d", subroom_id);
 
      Goal* wa = new WaitingArea();
```
zeroset: Hi, I hope you don't mind my suggestion for improvement. As i saw, the variable `wa` isn't used anywhere except for the return. My suggestion would be to create a `WaitingArea*` directly. With this you get rid of the dynamic cast. The return will automatically cast the `WaitingArea*` towards `Goal*` later on.   
```suggestion
     WaitingArea* waitingArea = new WaitingArea();
```

geometry/WaitingArea.cpp
```diff
@@ -1,93 +1,109 @@
...
+     return CheckProbabilities();
 }
 
-bool WaitingArea::checkProbabilities()
+bool WaitingArea::CheckProbabilities()
 {
      double p =0.;
-     for ( std::map<int, double>::iterator it = nextGoals.begin(); it != nextGoals.end(); it++ ){
-          p += it->second;
+     for ( auto it : _nextGoals ){
```
Ozaq: Could be rewritten with structured bindings:
```cpp
for(auto[key, value] : _nextGoals) {
    p += value;
}
// or
for(auto[_, value] : _nextGoals) {
    p += value;
}

File: geometry/WaitingArea.cpp [hunk 6]
```diff
@@ -99,118 +115,124 @@ std::string WaitingArea::toString()
...
      out.append(buffer);
 
-     for (auto const& nextWa : nextGoals){
+     for (auto const& nextWa : _nextGoals){
           sprintf(buffer, "\tnext id=%d\tp=%f\n", nextWa.first, nextWa.second);
           out.append(buffer);
      }
      out.append("]");
      return out;
-
```
Ozaq: Consider rewriting the whole function in C++:
```cpp
#incldue <sstream>
// ...
std::stringstream buffer;
buffer << "WaitingArea[\n\tid=" << _id 
           << "WaitingArea[\n\tcaption=" << _caption
//... and so on ... 
return buffer.str();

File: routing/RoutingEngine.h [hunk 7]
```diff
@@ -127,8 +112,8 @@ class RoutingEngine {
 private:
      /// collections of all routers used
      std::vector<Router*> _routersCollection;
-     /// collection of all trips/routes
-     Trips _tripsCollection;
+
+     /// states if the routers need to be updated
```
Ozaq: I think the filed name is already very good and does not need additional explanation with a comment

chraibi:
@schroedtert  some conflicts here. Wanna have a look and push again?

IO/GeoFileParser.cpp
```diff
@@ -715,61 +715,130 @@ Goal* GeoFileParser::parseGoalNode(TiXmlElement * e)
...
+
+     // Read is_open and check if correct value
+     if(const char* attribute = e->Attribute("is_open"); attribute) {
+          if (bool value = strcmp(xmltoa(attribute, "false"), "true") == 0) {
+               wa->SetOpen(value);
+          }
+     }
+
+     // Read global_timer and check if correct value
+     if(const char* attribute = e->Attribute("global_timer"); attribute) {
```
Ozaq: Same as for is_open. Also this attribute is not mentioned in the XSD, @chraibi is the XSD still updated?

IO/GeoFileParser.cpp
```diff
@@ -715,61 +715,130 @@ Goal* GeoFileParser::parseGoalNode(TiXmlElement * e)
...
-     waitingArea->SetOpen(open);
-     waitingArea->SetGlobalTimer(global_timer);
-     waitingArea->SetMinNumPed(min_peds);
-     waitingArea->SetMaxNumPed(max_peds);
-     waitingArea->SetWaitingTime(waiting_time);
-     waitingArea->SetTransitionID(transition_id);
-     waitingArea->SetRoomID(room_id);
-     waitingArea->SetSubRoomID(subroom_id);
+     // Additional checks:
+     bool waitPed = (wa->GetMinNumPed()  > 0 && wa->GetMaxNumPed() > 0 && wa->GetWaitingTime() > 0.);
```
Ozaq: should be 
```cpp
const bool waitPed = ...
```

IO/GeoFileParser.cpp
```diff
@@ -715,61 +715,130 @@ Goal* GeoFileParser::parseGoalNode(TiXmlElement * e)
...
+     bool waitDoor = (wa->GetTransitionID() > 0);
```
Ozaq: should be 
```cpp
const bool waitDoor = ...
```

Simulation.cpp
```diff
@@ -317,7 +317,7 @@ void Simulation::UpdateRoutesAndLocations()
 //    }
 
 #pragma omp parallel for shared(pedsToRemove, allRooms)
-     for (signed int p = 0; p < allPeds.size(); ++p) {
+     for (long unsigned int p = 0; p < allPeds.size(); ++p) {
```
Ozaq: This is probably better replaced by a range based for loop
```cpp
for(auto& ped : allPeds){
...
}
```

IO/GeoFileParser.cpp
```diff
@@ -715,61 +715,130 @@ Goal* GeoFileParser::parseGoalNode(TiXmlElement * e)
...
+
+     // Read waiting_time and check if correct value
+     if(const char* attribute = e->Attribute("waiting_time"); attribute) {
+          if (double value = xmltof(attribute, -1.); value>=0.) {
+               wa->SetWaitingTime(value);
+          }
+     }
+
+     // Read transition_id and check if correct value
+     if(const char* attribute = e->Attribute("transition_id"); attribute) {
```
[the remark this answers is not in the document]
schroedtert: The XSD is outdated, I’ll update it with the next commit. 
Following requirements need to be satisfied:
- Set `waiting_time`, `min_peds` and `max_peds`
OR
- Set `transition_id`
I’ve to figure out how to write this in the XSD file

Simulation.cpp
schroedtert: Apparently it does not work with clang and OpenMP 3.1 but compiles with gcc and OpenMP 4.5 (see Travis). From the standard I’m not entirely sure if it also distributes the computations across the cores or if it only compiles, since range based loop seem only be support with OpenMP 5.0

IO/GeoFileParser.cpp
```diff
@@ -723,56 +723,197 @@ Goal* GeoFileParser::parseGoalNode(TiXmlElement * e)
...
+          return nullptr;
+     }
 
-     std::map<int, double> nextGoals;
+     // Read caption and check if correct value
+     if(const char* attribute = e->Attribute("caption"); attribute) {
+          if (std::string value = xmltoa(attribute, ""); !value.empty()) {
+               wa->SetCaption(value);
+          }else{
+               wa->SetCaption("WA " + std::to_string(wa->GetId()));
```
Ozaq: Caption is still required in the updated XSD, this code however will be ok with a missing caption

Simulation.cpp
Ozaq: I see, I did also some googling and it seems that it should (TM) work if it passes the compiler but I could not find a clear answer :/ Good call to leave it at this form. @chraibi Should there be a min OpenMP version specified?

chraibi: unless there is a very good reason for requiring a minimum OpenMP version, I would not do it.

IO/GeoFileParser.cpp
schroedtert: Removed the required from the XSD

chraibi:
@Ozaq still almost there?

schroedtert:
@Ozaq Anything else which should be considered?

chraibi:
@Ozaq I would merge this PR if you don't have any more suggestions. Please give me a thump up if we are ready to go.

Ozaq:
> @Ozaq Anything else which should be considered?

Sorry, I didn't see that message :/

PR #421, opened 2019-07-18
Fix compile error on clang

Ozaq:
Last falltrough statement was misplaced and caused a compile error on
clang. [[falltrough]] has to precede the case to continue to.


chraibi:
@Ozaq 

I can't compile with clang++-8  on linux. [see this build](https://travis-ci.com/chraibi/jpscore/jobs/217457714)

Is your CMakeLists.txt different than mine or do you use a higher clang-version?

PR #424, opened 2019-07-19
Travis

chraibi:
Build fail, especially for clang++.
Some tests fail too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,68 +1,27 @@
 language: cpp
-dist: xenial
+dist: bionic
+
 git:
-  depth: 5
+  depth: 1
   quiet: true
-
-python: 3.7
-
-
 
 matrix:
   include:
-  # todo
-    # - name: Static checks (clang-format)
-    #   env: STATIC_CHECKS=yes
-    #   os: linux
-    #   compiler: gcc
-    #   addons:
-    #     apt:
-    #       sources:
-    #         - llvm-toolchain-xenial-8
-    #       packages:
-    #         - clang-format-8
-    #   script:
-    #    - $CXX --version
-    #    - mkdir build && cd build
-    #    - cmake -DBUILD_TESTING=ON ..
-    #    - make
-
-    # todo
-    # - name: osx
-    #   os: osx
-    #   osx_image: xcode11
-    #   compiler: clang
-    #   script:
-    #    - mkdir build && cd build
-    #    - cmake  ..
-    #    - make
-
     - name: linux (clang)
       os: linux
       compiler: clang++
       script:
        - mkdir build && cd build
        - cmake  ..
-       - make
-    # run tests only here
+       - make -j $(nproc)
+
     - name: linux (gcc)
       os: linux
       compiler: gcc
       script:
-       - $CXX --version
        - mkdir build && cd build
-       - cmake -DBUILD_TESTING=ON ..
-       - make
-       - ctest --verbose -R dev
-       - ctest -R clean
-       - ctest --verbose -R rimea
-#      - ctest -R juelich
-
-    # todo
-    # for windows boost is too old. Chocolatey is outdated
-    # - name: windows
-    #   os: windows
-
+       - cmake ..
+       - make -j $(nproc)
 
 before_script:
     - source .travis_before_script.bash
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,74 @@
+language: cpp
+dist: xenial
+git:
+  depth: 5
+  quiet: true
+
+python: 3.7
+
+
+
+matrix:
+  include:
+  # todo
+    # - name: Static checks (clang-format)
+    #   env: STATIC_CHECKS=yes
+    #   os: linux
+    #   compiler: gcc
+    #   addons:
+    #     apt:
+    #       sources:
+    #         - llvm-toolchain-xenial-8
+    #       packages:
+    #         - clang-format-8
+    #   script:
+    #    - $CXX --version
+    #    - mkdir build && cd build
+    #    - cmake -DBUILD_TESTING=ON ..
+    #    - make
+
+    # todo
+    # - name: osx
+    #   os: osx
+    #   osx_image: xcode11
+    #   compiler: clang
+    #   script:
+    #    - mkdir build && cd build
+    #    - cmake  ..
+    #    - make
+
+    - name: linux (clang)
+      os: linux
+      compiler: clang++
+      script:
+       - mkdir build && cd build
+       - cmake  ..
+       - make
+    # run tests only here
+    - name: linux (gcc)
+      os: linux
+      compiler: gcc
+      script:
+       - $CXX --version
+       - mkdir build && cd build
+       - cmake -DBUILD_TESTING=ON ..
+       - make
+       - ctest --verbose -R dev
+       - ctest -R clean
+       - ctest --verbose -R rimea
+#      - ctest -R juelich
+
+    # todo
+    # for windows boost is too old. Chocolatey is outdated
+    # - name: windows
+    #   os: windows
+
+
+before_script:
+    - source .travis_before_script.bash
+
+before_install:
+    - source .travis_before_install.bash
+
+install:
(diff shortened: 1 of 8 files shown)


Ozaq:
I had a look at the build issues in Travis:

About the failing test: https://travis-ci.org/JuPedSim/jpscore/jobs/560924513
The failing tests is due to a timeout on Travis side. The tests run about 30-40x slower than on my local machine and `test_clean_geometry_7` takes aprox. 30sec locally, i.e. the test runs somewhat between 15-20 minutes in Travis. Given the error message in Travis: 

```
No output has been received in the last 10m0s, this potentially indicates a stalled build 
or something wrong with the build itself. 
```
This message clarifies that Travis aborted the test due to not producing any output in the expected time frame.

There is a possible workaround however, you could run as `ctest -V -R clean` instead which puts ctest in a mode where it dumps all output to the console. Still the testing time is atrocious :/ 

About the build error: https://travis-ci.org/JuPedSim/jpscore/jobs/560924512
What is happening here is that clang always defaults to use the standard library implementation from gcc (otherwise there would be incompatibilities with system packages, if the system packages expose types like std::string in their interface). That means that clang 7 here is using GCC 5.4 (default on Ubuntu 16.04) standard library which does not yet support filesystem. Filesystem was first available in GCC 8 so you would need to install GCC 8 and probably also update alternatives to make it the default compiler when calling `gcc` or `g++`.

.travis_before_install.bash
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]
+then
+    if [ "$CXX" == "g++" ]
```
Ozaq: Both clang and GCC builds need the newer libstdc++ to have filesystem support so GCC 9 should be installed unconditionally 
```suggestion
    sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
    if [ "$CXX" == "clang++" ]
    then
        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
        sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main"
    fi
```

.travis_install.bash
```diff
@@ -0,0 +1,17 @@
...
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]
+then
+    if [ "$CXX" = "g++" ];
+    then
+        sudo apt-get install -qq g++-9
+        export CXX="g++-9"
+    elif [ "$CXX" == "clang++" ]
+    then
+        sudo apt-get install -y  --allow-unauthenticated -qq clang-8 libomp-8-dev libc++-8-dev libc++abi-8-dev
```
Ozaq:
```suggestion
        sudo apt-get install -y --allow-unauthenticated -qq clang-8 libomp-8-dev g++-9
```
Since you are by default not using libc++ but rather libstdc+ you don't need those, instead add gcc-9 for a libstdc++ that supports filesystem.

Also: Consider removing `--allow-unauthenticated`, gpg-keys have been installed this really should work w.o. this flag

CMakeLists.txt
```diff
@@ -664,14 +654,17 @@ if(JPSFIRE)
 endif()
 
 # todo make this system
-add_library(thirdParty STATIC ${THIRD_PARTY_SRC})
+add_library(thirdParty SHARED ${THIRD_PARTY_SRC})
 
 add_library(core STATIC ${source_files})
 if(CMAKE_COMPILER_IS_GNUCXX)
```
Ozaq: 1st remark: cmake doc strongly suggest to not use `CMAKE_COMPILER_IS_GNUCXX` but instead use `${CMAKE_CXX_COMPILER_ID} EQUALS "GNU"` instead.

2nd remark clang and gcc link libs are now identical, you can say:
```cmake
if(${CMAKE_CXX_COMPILER_ID} MATCHES "^(GNU|Clang)$")
```

.travis.yml
```diff
@@ -0,0 +1,74 @@
+language: cpp
+dist: xenial
+git:
+  depth: 5
```
Ozaq: Why not 1?

.travis.yml
```diff
@@ -0,0 +1,74 @@
...
+  quiet: true
+
+python: 3.7
+
+
+
+matrix:
+  include:
+  # todo
+    # - name: Static checks (clang-format)
```
Ozaq: I would suggest to leave out the commented out part and issue a separate PR where this all is implemented.

.travis.yml
```diff
@@ -0,0 +1,74 @@
...
+    #    - cmake  ..
+    #    - make
+
+    - name: linux (clang)
+      os: linux
+      compiler: clang++
+      script:
+       - mkdir build && cd build
+       - cmake  ..
+       - make
```
Ozaq: Travis supplies 2 cores to each VM so you can run this with make -j $(nproc) to get a little bit more speed

.travis.yml
```diff
@@ -0,0 +1,74 @@
...
+    # run tests only here
+    - name: linux (gcc)
+      os: linux
+      compiler: gcc
+      script:
+       - $CXX --version
```
Ozaq: You could leave that out since CMake will dump the compiler found to stdout

CMakeLists.txt
```diff
@@ -664,14 +654,17 @@ if(JPSFIRE)
 endif()
 
 # todo make this system
-add_library(thirdParty STATIC ${THIRD_PARTY_SRC})
```
Ozaq: I don't think it is necessary to change to dynamic linking

chraibi:
> The failing tests is due to a timeout on Travis side. The tests run about 30-40x slower than on my local machine and `test_clean_geometry_7` takes aprox. 30sec locally, i.e. the test runs somewhat between 15-20 minutes in Travis. Given the error message in Travis:

On my machine, the tests succeed too. This probably means that we gotta run the heavy tests on a [local server](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/pipelines/22454). I did not even activate the rimea-tests, which take way too much time.

> About the build error: https://travis-ci.org/JuPedSim/jpscore/jobs/560924512
> What is happening here is that clang always defaults to use the standard library implementation from gcc (otherwise there would be incompatibilities with system packages, if the system packages expose types like std::string in their interface). That means that clang 7 here is using GCC 5.4 (default on Ubuntu 16.04) standard library which does not yet support filesystem. Filesystem was first available in GCC 8 so you would need to install GCC 8 and probably also update alternatives to make it the default compiler when calling `gcc` or `g++`.

This is bizarre.  For the sake of testing and understanding this behavior I use this [project](https://travis-ci.com/chraibi/c-17fs/jobs/217619411)

which works on a local ubuntu image, but not with travis. Although, on travis clang++-8 is used 

if `clang` uses `g++` (I'm not sure I understand this), then what is the point to test compilation with both compilers?

.travis.yml
```diff
@@ -0,0 +1,74 @@
+language: cpp
+dist: xenial
```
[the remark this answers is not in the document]
chraibi: makes sense to use `bionic`. Thanks!

.travis.yml
chraibi: 1 would be enough.

.travis.yml
```diff
@@ -0,0 +1,74 @@
...
+    - name: linux (gcc)
+      os: linux
+      compiler: gcc
+      script:
+       - $CXX --version
+       - mkdir build && cd build
+       - cmake -DBUILD_TESTING=ON ..
+       - make
+       - ctest --verbose -R dev
+       - ctest -R clean
```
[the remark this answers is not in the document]
chraibi: I removed all these tests due to the heavy computation cost until we figure out a better solution

.travis_install.bash
chraibi: makes sense. Thanks!

CMakeLists.txt
chraibi: That was not intentional. `Static` should do.

PR #435, opened 2019-07-20
Fix build error when enableing JPSFIRE

Ozaq:
Please have a look at this one: I replaced the faulty return statements with `std::exit(EXIT_FAILURE)` since it looked like this was the actual intention. I did NOT run any tests on this one. JPSFIRE seems to be in a bad state. I just wanted to check for compile issues (which i found) and additional warnings (which there are none, except cnpy)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,28 +7,27 @@
 # -DUSE_DUAL_ABI=ON (default OFF)  https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
 # -D Boost_NO_SYSTEM_PATHS=true (default false) -D BOOST_ROOT=PATH_where_to_find_boost
 # -D AIROUTER=true (default false)
-# -D JPSFIRE=true (default false)
+# -D JPSFIRE=ON (default OFF)
 #--------------------------------------------------------------------------
-
+################################################################################
+# Project setup
+################################################################################
 cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -7,30 +7,29 @@
...
-
-# Project setup
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+################################################################################
+# Optional features
+################################################################################
+set(JPSFIRE OFF CACHE BOOL "Build with jps fire support")
```
chraibi: this replaces the if else block?
Then maybe it makes sense to change the  AIROUTER block as well!

Ozaq: This replaces the if else yes. I did not include this since this PR was just about JPSFIRE. I have a few more cleanups for the CMake file locally ready. How about Ill include the change for the AIROUTER in an upcoming PR that just deals with CMake cleanup?

chraibi: makes sense. Could you delete this [line](https://github.com/Ozaq/jpscore/blob/get_jpsfire_compiling_again/CMakeLists.txt#L634)?

Ozaq: Done, also moved the 2nd output of the JSPFIRE status up to the declaration to keep it "closer together"

chraibi:
I run this with this [demo](https://gitlab.version.fz-juelich.de/jupedsim/jpsfire/tree/master/demos/0_demo).  Looks good! Thanks!

Issue #412, opened 2019-07-16
Reseting max_agents 


anna-braun:
In _traffic_constraints_ we can give doors the parameter _max_agents_. If _max_agents_ have passed a door it will be closed permanently. For several reasons it should be possible to reopen such doors and reset the counter of _max_agents_ when the doors are reopened. Here are two ideas for it:     
1.  Adapting the schedules so that they can be controlled by time and by max_agents.      
-> The doors will be closed after a defined time or when a maximum number of agents have passed the doors. The counter for the _max_agents_ should be reseted when the door reopens after the closing time:
```xml 
<time group_id="0" closing_time="20" max_agents="300">
    <t t="30"/>
    <t t="80"/>
</time>
```
2. Reset the counter of max_agents if the door is opened by an event.      
-> Set the counter of _max_agents_ in _traffic_constraints_ to zero when the door is opened by an event:

chraibi:
I guess solution 2 is more intuitive. 

The original purpose of `max_agents` was to assure a (*permanent*) limit for the door. What if you simply take a very big value (inf)?

schroedtert:
@chraibi Is the regulation of the flow of normal still working? For a simple case in the current commit 6b9e0f0b2d4469ecdcc326937e0d7f03fdf5cced it did not work for me. The is not closed after max_agents passed through it.

[412-reset-door.zip](https://github.com/JuPedSim/jpscore/files/3412113/412-reset-door.zip)

chraibi:
It works fine, but, you use `dn=0`. 

However, according to http://www.jupedsim.org/jpscore/2016-11-01-inifile.html#traffic-constraints

> `dn:` number of agents to pass the door before triggering the process of flow regulation.

In [simulation.cpp](https://github.com/JuPedSim/jpscore/blob/develop/Simulation.cpp#L1112)
```
                 if( trans->GetPartialDoorUsage() ==  trans->GetDN() ) {
                      trans->regulateFlow(Pedestrian::GetGlobalTime());
                      trans->ResetPartialDoorUsage();
                 }
```
the condition is never triggered when `dn==0`.

Another *hack*: 

This all works only when you activate 

Because as we can see in this  [line](https://github.com/JuPedSim/jpscore/blob/develop/Simulation.cpp#L1069)

`Simulation::UpdateFlowAtDoors` is called when this conditio is true
```
    if (_config->ShowStatistics())
```

To summarize:  It works with

```
 <show_statistics>true</show_statistics> 
[...]

PR #438, opened 2019-07-21
Statistics

chraibi:
This PR is related to a trap mentioned in my comment in #412

### DIFF
--- a/IO/GeoFileParser.cpp
+++ b/IO/GeoFileParser.cpp
@@ -496,7 +496,7 @@ bool GeoFileParser::parseDoorNode(TiXmlElement * xDoor, int id, Building* buildi
      }
      //----------------- dn
      int DN = xmltof(xDoor->Attribute("dn"), -1.0);
-     if(DN >= 0)
+     if(DN > 0)
      {
           building->GetTransition(id)->SetDN(DN);
           sprintf(tmp, "\t>> dn: %d\n", DN);
--- a/Simulation.cpp
+++ b/Simulation.cpp
@@ -711,8 +711,10 @@ double Simulation::RunBody(double maxSimTime)
         }// Transitions
         if(frameNr % 1000 == 0)
         {
-              Log->Write("INFO:\tUpdate door statistics at t=%.2f", t);
-              PrintStatistics(t);
+             if (_config->ShowStatistics()){
+                  Log->Write("INFO:\tUpdate door statistics at t=%.2f", t);
+                  PrintStatistics(t);
+             }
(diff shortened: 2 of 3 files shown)


IO/GeoFileParser.cpp
```diff
@@ -496,7 +496,7 @@ bool GeoFileParser::parseDoorNode(TiXmlElement * xDoor, int id, Building* buildi
      }
      //----------------- dn
      int DN = xmltof(xDoor->Attribute("dn"), -1.0);
-     if(DN >= 0)
+     if(DN > 0)
```
schroedtert: With this, it is not possible to start the counting right at the beginning of the simulation, right? For example if I want to let 5 peds through a door: dn = 0 and max_agents = 5 won’t work

chraibi: The decisive condition is in this [line](https://github.com/JuPedSim/jpscore/blob/develop/Simulation.cpp#L1112)

```
if( trans->GetPartialDoorUsage() == trans->GetDN() )
```
`PartialDoorUsage is > 0` because in the [line before](https://github.com/JuPedSim/jpscore/blob/develop/Simulation.cpp#L1106)
it's incremented by 1.

So `dn>0` should be satisfied otherwise that if will not be entered.

schroedtert: Basically my question is, if it will be possible to start the flow regulation with the beginning of the simulation or only if x>=1 peds have passed the door?

chraibi: The regulation of a transition kicks in once at least 1 pedestrian passes that transition. 

The reason why we introduced `dn` was to avoid too many updates.  

So, a reasonable value for `dn` should definitely depend on the simulation size.

`dn=0` makes no sense in this regard. `dn=1`could lead to too frequent updates.

Maybe, it would be more user friendly if we remove this parameter and decide for a default value. But, then we could lose some flexibility too.

schroedtert: Just for my understanding: If I set `dn=5` and `max_agents=5`, will the door closed after 5 or 10 agents passed it?

chraibi: In that case, 5 pedestrians will pass the door. 
```
if( trans->GetPartialDoorUsage() == trans->GetDN() )

```

is true 

--> Regulation starts. 
But then `max_agents=5` 
--> door closes permanently!

schroedtert: Ok, then I understood that part wrong.

PR #440, opened 2019-07-21
Cleanup cmake part2

Ozaq:
So I have been working on cleaning up the cmake file. I would like you to already have a look at the changes. Quite a lot has accumulated already.

Currently there are some caveats:
1.) Semantics of `BUILD_TESTING` and `BUILD_CPPUINT_TEST` have slightly changed. `BUILD_TESTING` now controls integration tests only and `BUILD_CPPUNIT_TESTING` controls unit tests only

2.) I have no windows machine available atm so I did not check for any breakage on windows. I probably will add a travis target for windows at some point before merging

if you already spot a no-go let me know :)

PR #441, opened 2019-07-21
412 reseting max agents

schroedtert:
Updates the event mechanism to also allow resetting the door usage.


geometry/Crossing.cpp
```diff
@@ -226,6 +227,17 @@ int Crossing::GetPartialDoorUsage() const
      return _partialDoorUsage;
 }
 
+void Crossing::ResetDoorUsage()
+{
+     _doorUsage = 0;
+
+     if ((_outflowRate >= std::numeric_limits<double>::max())){
```
chraibi: you mean `<`?

schroedtert: No, the door flow should only be closed, if no outflowRate set but a maxDoorUsage

Simulation.cpp
```diff
@@ -358,7 +358,7 @@ void Simulation::UpdateRoutesAndLocations()
 //    }
 
 #pragma omp parallel for shared(pedsToRemove, allRooms)
-     for (long unsigned int p = 0; p < allPeds.size(); ++p) {
+     for (size_t p = 0; p < allPeds.size(); ++p) {
```
chraibi: there is a little conflict here with @xuqiancheng. 

Apparently, in windows, this change does not work. See also [here](https://github.com/JuPedSim/jpscore/pull/442/commits/cd0571e5b586ef22e5abfe39abdbaf8d03e33dd3#diff-441e5250dd8ecdf163d639bb5ac99da4L453)

schroedtert: So should we stay with int? Or should we wait till #440 is merged and see if this solves this?

Ozaq: @chraibi The only reason i can think of why this does not work is if the library type `size_t` is not "visible" i.e. the STD headers on linux / windows differ a bit in what they expose coincidentally. Looking at `Simulation.h` and 'Simulation.cpp` I see that there is no include `std::map` or `std::shared_ptr` so even those types are only coincidentally visible. It is considered essential practice in C++ to include all types a cpp/h file needs explicitly and not rely on other headers to transitively supply the type.

To fix the size_t issue on windows (and in general)  do in simulation.cpp:
```cpp
#include <cstddef>
```

Issue #420, opened 2019-07-18
Pedestrian Distribution


planospeedo:
in function **_PossiblePositions_** from **_PedDistributor.cpp_**

`

    double bufx = 0.12;
    double bufy = 0.12;
    double amin = 0.18; // = GetAmin()->GetMean();
    double bmax = 0.25; // = GetBmax()->GetMean();
    double dx = amin + bufx;
    double dy = bmax + bufy;

`

With this, the distribution is hardcoded to dx = 0.3 m and dy = 0.37 m, which is a problem when using bigger pedestrian sizes and higher densities as overlapping initial distributions and no movement (at least when using the velocity model) occurs.

Example: pedestrian size is 0.4 m (l=0.2m)
First image: overlapping as amin=0.18

![Bildschirmfoto 2019-07-18 um 16 12 16](https://user-images.githubusercontent.com/15627872/61467368-c6081200-a97b-11e9-807e-5e680e5e2a28.png)

Second image: no overlapping when setting amin=0.25 for instance

![Bildschirmfoto 2019-07-18 um 16 13 13](https://user-images.githubusercontent.com/15627872/61467377-c86a6c00-a97b-11e9-95c1-049155508aca.png)

gjaeger:
@planospeedo Should the values (in the code) only be increased? Or should they be dependent on the input?

planospeedo:
@gjaeger input-dependent I suppose, otherwise it is for instance not possible to have high initial densities with smaller pedestrian sizes

gjaeger:
OK. That would be my suggestion too.

gjaeger:
The change was made at commit c30e7ad on Jul 4, 2014.

gjaeger:
@planospeedo can wie close this issue?

PR #442, opened 2019-07-22
 initial distribution of agents depends on the size of agents

xuqiancheng:
#420
Now initial distribution of agents depends on the size of agents.
But when the size of agents is normal distribution instead of a constant value, overlapping in initial distribution can still be observed.
May be we can introduce something like buffer distance between agents when doing initial distributing to avoid overlapping.  

Also make some changes to compile jpscore on windows with VS2017:
- VS2017 use OpenMP 2.0, so index variable in OpenMP 'for' statement must have signed integral type
- Add some code in CmakeLists.txt files for using MSVC
- The input argument of function 'void Configuration::SetTrajectoriesFile(std::string)' should be a std::sting instead of std::filesystem::path


pedestrian/AgentsParameters.cpp
```diff
@@ -432,5 +432,13 @@ std::string AgentsParameters::writeParameter()
     return s;
 }
 
+double AgentsParameters::GetAminMean() {
+	return _Amin.mean();
```
planospeedo: spaces not tabs

pedestrian/AgentsParameters.cpp
```diff
@@ -432,5 +432,13 @@ std::string AgentsParameters::writeParameter()
...
+}
+
+double AgentsParameters::GetBmaxMean() {
+	return _Bmax.mean();
```
planospeedo: spaces not tabs

pedestrian/AgentsSourcesManager.cpp
```diff
@@ -214,7 +214,9 @@ void AgentsSourcesManager::ComputeBestPositionCompleteRandom(AgentsSource* src,
 {
      auto dist = src->GetStartDistribution();
      auto subroom = _building->GetRoom(dist->GetRoomId())->GetSubRoom(dist->GetSubroomID());
-     vector<Point> positions = PedDistributor::PossiblePositions(*subroom);
+     auto config=GetBuilding()->GetConfig();
+     auto distributor = std::unique_ptr<PedDistributor>(new PedDistributor(config));
+     vector<Point> positions = distributor->PossiblePositions(*subroom);
```
planospeedo: Why was this changed?

xuqiancheng: > Why was this changed?

Because I have changed the function PossiblePosition from static to non-static.
May be there are better way to doi it.

PR #447, opened 2019-07-30
General cleanup

Ozaq:
Looking at the `size_t` Issue from another PR i though Ill do some cleanup. This is a first PR of probably a few more.

### DIFF
--- a/general/OpenMP.h
+++ b/general/OpenMP.h
@@ -1,8 +1,6 @@
 /**
- * \file        ForceModel.cpp
- * \date        Dec 13, 2010
- * \version     v0.7
- * \copyright   <2009-2015> Forschungszentrum Jülich GmbH. All rights reserved.
+ * \version     v0.8
+ * \copyright   <2009-2019> Forschungszentrum J?lich GmbH. All rights reserved.
  *
  * \section License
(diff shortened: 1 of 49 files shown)


general/OpenMP.h
```diff
@@ -0,0 +1,35 @@
...
+ * along with JuPedSim. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * \section Description
+ * The Simulation class represents a simulation of pedestrians
+ * based on a certain model in a specific scenario. A simulation is defined by
+ * various parameters and functions.
+ *
+ *
+ **/
+#pragma once
```
chraibi: what is the meaning of this line?

Ozaq: Its a replacement for include guards:
```cpp
#indef FOO_H
#define FOO_H
... your code ...
#endif
```
can be replaced by a simple 
```cpp
#pragma once
```
it has 2 small benefits:
a) there can never be an name collision as with include guards. Imagine a situation where two include files use the same define as include guards. You could not include those headers in any file together.
b) single line vs 3 lines, less verbose

chraibi: that's interesting! Thanks!

PR #460, opened 2019-08-08
Make all paths typesafe by using fs::path

Ozaq:
When trying to run `demos/scenario_3_corner` I encountered an issue when `jpscore` was trying to read `corner_routing.xml`. `jpscore` was missing a trailing `/` on the parent path so that the last part of the path and the filename got melded into one aka `scenario_3_cornercorner_routing.xml`.

Since this seems to be common and prevailing issue I am refactoring all use of path / file identifiers to use fs::path instead of strings so that paths can be safely and correctly concatenated on all operating systems.

I am opening the PR since I ~might~ need help to rack down all references to files / paths and I know this not yet complete. Hints where to find references to files are very welcome!

### DIFF
--- a/routing/global_shortest/GlobalRouter.cpp
+++ b/routing/global_shortest/GlobalRouter.cpp
@@ -1331,7 +1331,7 @@ bool GlobalRouter::GenerateNavigationMesh()
 fs::path GlobalRouter::GetRoutingInfoFile()
 {
 
-     TiXmlDocument doc(_building->GetProjectFilename().c_str());
+     TiXmlDocument doc(_building->GetProjectFilename().string());
      if (!doc.LoadFile()) {
           Log->Write("ERROR: \t%s", doc.ErrorDesc());
           Log->Write("ERROR: \t GlobalRouter: could not parse the project file");
@@ -1411,7 +1411,7 @@ bool GlobalRouter::LoadRoutingInfos(const fs::path &filename)
      Log->Write("INFO:\tLoading extra routing information for the global/quickest path router");
      Log->Write("INFO:\t  from the file "+filename.string());
 
-     TiXmlDocument docRouting(filename.c_str());
+     TiXmlDocument docRouting(filename.string());
      if (!docRouting.LoadFile()) {
           Log->Write("ERROR: \t%s", docRouting.ErrorDesc());
           Log->Write("ERROR: \t could not parse the routing file [%s]",filename.c_str());
--- a/routing/global_shortest/GlobalRouter.cpp
+++ b/routing/global_shortest/GlobalRouter.cpp
@@ -1328,10 +1328,10 @@ bool GlobalRouter::GenerateNavigationMesh()
      return true;
 }
 
-std::string GlobalRouter::GetRoutingInfoFile()
+fs::path GlobalRouter::GetRoutingInfoFile()
 {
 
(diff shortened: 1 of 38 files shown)


chraibi:
This is a nice PR and really great efforts. Thanks @Ozaq 

I believe it's still not ready yet. 
Tested on a mac. Here are the logs:

[cmake_log.txt](https://github.com/JuPedSim/jpscore/files/3495390/cmake_log.txt)
[ninja_log.txt](https://github.com/JuPedSim/jpscore/files/3495392/ninja_log.txt)

Ozaq:
@chraibi thx. About the build errors: turns out that boost filesystem does not implicitly convert to string or char* so `.string()` and `.c_str()` calls have to be added.

Ozaq:
@chraibi commit is updated with missing calls, this should(TM) work now ;) @xuqiancheng would you be son kind and give this a try on windows?

xuqiancheng:
@Ozaq 
I try to compile jpcore on windows, but failed. 
I have met some errors as shown in the figure.
It seems that the function std::filesystem::path::c_str() doesn't work. 
image

Ozaq:
@xuqiancheng thanks for testing... The problem here is that by default on windows filesystem::path is based on wchar and `c_str()` returns `wchar*`. I can fix that just that I now have to replace all calls to `c_str()` with `string().c_str()`... fun times...

chraibi:
I get similar errors when compiling with AppleClang too.

Ozaq:
> I get similar errors when compiling with AppleClang too.

Do you get those errors on the latest change as well? It compiles w.o. errors on my Mac now. I did force push over the commit so you need to reset to the changes on origin.

I can address the Windows errors tonight but I am a bit caught up at work atm so I am a bit slow to respond.

Ozaq:
@xuqiancheng would you mind to retest on the latest commit

routing/global_shortest/GlobalRouter.cpp
```diff
@@ -1400,18 +1400,18 @@ std::string GlobalRouter::GetRoutingInfoFile()
...
 {
-     if(filename=="") return true;
+     if(filename.empty()) return true;
 
      Log->Write("INFO:\tLoading extra routing information for the global/quickest path router");
-     Log->Write("INFO:\t  from the file "+filename);
+     Log->Write("INFO:\t  from the file "+filename.string());
 
-     TiXmlDocument docRouting(filename);
+     TiXmlDocument docRouting(filename.c_str());
```
xuqiancheng: @Ozaq Many thanks for these work !

- I got an error here , and after changing to filename.string(), I compile jpscore on windows successfully. I have tested demos/scenario_3_corner, it works. 

- And there is another small problem, which is the location of geometry file in trajectory file is wrong. I have attached a screenshot here, I think it's easy to fix.

image

Ozaq: Thanks for testing, ill have a look

Ozaq: @xuqiancheng It should(TM) work now

xuqiancheng: @Ozaq Yes, it works now !

chraibi:
Thanks, @Ozaq  for this PR.

PR #462, opened 2019-08-08
Add option to build with tsan

Ozaq:
New cmake option added BUILD_WITH_TSAN (Default=OFF), builds a new
target jpscore_tsan.

Also updates asan target with undefined behaviour sanitizer, since ubsan
and asan can work in conjunction.


Ozaq:
*sigh* for this to be even remotly useful libomp would need to be compiled with `-fsanitize=thread` as well...

Issue #476, opened 2019-08-14
vtk file function


gjaeger:
**JuPedSim - JPScore**

Current date   : Aug 14 2019 10:20:22
Version        : 0.8.4
Compiler       : clang++ (10.0.0)
Commit hash    : v0.8.4-164-g082762a9
Commit date    : Wed Aug 14 08:43:29 2019
Branch         : develop

my example: [ini_UNI_CORR_500_05_om_3.xml.zip](https://github.com/JuPedSim/jpscore/files/3500479/ini_UNI_CORR_500_05_om_3.xml.zip) and [geo_UNI_CORR_500_05_jpscore.xml.zip](https://github.com/JuPedSim/jpscore/files/3500481/geo_UNI_CORR_500_05_jpscore.xml.zip)

**Describe the bug**
vtk-files can be written with `<write_VTK_files>true</write_VTK_files>` since [v0.8.3](https://github.com/JuPedSim/jpscore/blob/develop/CHANGELOG.md#v083-2018-05-18).

With this setting 
`<write_VTK_files>false</write_VTK_files>`
the file will still be created.

My expectation is that the file will not be created.

gjaeger:
``` 
JuPedSim - JPScore

Current date   : Aug 18 2019 18:37:43
Version        : 0.8.4
Commit hash    : v0.8.4-165-g41882995
Commit date    : Sun Aug 18 14:44:48 2019
Branch         : pr478
```
The element  `<write_VTK_files>false</write_VTK_files>` works as desired.

Unfortunately the input `true` does not generate any output (`goal.vtk`).

PR #478, opened 2019-08-18
Fix vtx export

Ozaq:
jpscore was writing goals.vtk file even though writing vtk files was
disabled.

Fixes: #476

### DIFF
--- a/routing/ff_router/ffRouter.cpp
+++ b/routing/ff_router/ffRouter.cpp
@@ -106,7 +106,9 @@ bool FFRouter::Init(Building* building)
           _goalToLineUIDmap = _globalFF->getGoalToLineUIDmap();
           _goalToLineUIDmap2 = _globalFF->getGoalToLineUIDmap2();
           _goalToLineUIDmap3 = _globalFF->getGoalToLineUIDmap3();
-          _globalFF->writeGoalFF("goal.vtk", goalIDs);
+          if(_building->GetConfig()->get_write_VTK_files()) {
+              _globalFF->writeGoalFF("goal.vtk", goalIDs);
+          }
      }
      //get all door UIDs
      _allDoorUIDs.clear();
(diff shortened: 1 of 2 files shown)


routing/ff_router/ffRouter.cpp
```diff
@@ -106,7 +106,9 @@ bool FFRouter::Init(Building* building)
           _goalToLineUIDmap = _globalFF->getGoalToLineUIDmap();
           _goalToLineUIDmap2 = _globalFF->getGoalToLineUIDmap2();
           _goalToLineUIDmap3 = _globalFF->getGoalToLineUIDmap3();
-          _globalFF->writeGoalFF("goal.vtk", goalIDs);
+          if(_building->GetConfig()->get_write_VTK_files()) {
+              _globalFF->writeGoalFF("goal.vtk", goalIDs);
```
gjaeger: The element <write_VTK_files>false</write_VTK_files> works as desired.

Unfortunately the input true does not generate any output (goal.vtk).

Ozaq: Hmm strange... Ill look into it

Ozaq: Ok it looks like `write_VTK_files` has to be an element under FFRouter params, e.g.:
```xml
<route_choice_models>
  <router router_id="1" description="ff_global_shortest">
    <parameters>
      <write_VTK_files>true</write_VTK_files>
    </parameters>
  </router>
</route_choice_models>
```

Additionally the same XML element is reused under 
```xml
<operational_models>
  <model operational_model_id="3" description="Tordeux2015">
    <model_parameters>
      <write_VTK_files>true</write_VTK_files>
    </model_parameters>
```
But here this sets `write_vtk_direction` on config.

@chraibi Can you help me out here? Is this the desired behavior or is there more fix?

zeroset: `write_VTK_files` is a Parameter of the FFRouters and therefore should be part of the parameters of the router. 
Additionally I created a new issue for renaming the model specific `write_VTK_files` parameter (#484).

Issue #310, opened 2019-07-11
Logger


chraibi:
In Gitlab by @schroedtert on May 6, 2019, 09:44 [[origin](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/issues/317)]

Issue for collecting ideas for logging libraries

chraibi:
In Gitlab by @chraibi on May 10, 2019, 15:35

In my opinion, it should be
- a header only library
- works throughout all files and write logs in a file

An example: 

https://github.com/gabime/spdlog

zeroset:
We use SPDLog.

PR #479, opened 2019-08-18
Spdlog integration

Ozaq:
Currently logging uses char* and format strings all over the place resulting in issues with filesystem::path to cstring conversion on windows.

Since there is already #310 I though Ill have a quick experiment on how a switch to spdlog might look like.

Is this change something you would like to have?

### DIFF
--- a/third-party/CMakeLists.txt
+++ b/third-party/CMakeLists.txt
@@ -49,6 +49,7 @@ target_include_directories(visilibity PUBLIC
 # Build poly2tri
 ################################################################################
 add_library(poly2tri STATIC
+
     poly2tri/common/shapes.cpp
     poly2tri/common/shapes.h
     poly2tri/common/utils.h
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,16 +11,18 @@ matrix:
       os: linux
       env: TOOLCHAIN=clang
       script:
+       - ./scripts/setup-deps.sh
        - mkdir build && cd build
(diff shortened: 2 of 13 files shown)


Ozaq:
Hi @zeroset did you have a chance to look into this yet? Is this of interest? If yes let me know how we should proceed. Cheers!

zeroset:
Hi, i really much like the implementation especially in combination with libfmt. I am actually not sure if we should use both libraries as a git submodule or better add a library requirement as a system library. But submodule seems to be more convenient.

I would give it a YES, but nevertheless, let's wait for an opinion from @chraibi 
Thanks!

chraibi:
> Is this change something you would like to have?

Definitely yes.  👍 

@Ozaq:  We had this before, I think it is still important to dumb logs in a logfile *and* write out to the screen.

chraibi:
@Ozaq  Sorry for being late, but I do love this PR. I'll check it and come back to you.

chraibi:
Screenshot 2019-09-06 13 06 39

Ozaq:
> To `cmake -DUSE_OPENMP=OFF ..` I get the following errors.
> Am I supposed to download something before?
> 
> Screenshot 2019-09-06 13 06 39

Woops sorry for not getting back to you, wrote an answer but never hit send... Eh yeah you would need to update the submodules but since then I reworked the PR to use system dependencies.

third-party/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ target_include_directories(visilibity PUBLIC
 # Build poly2tri
 ################################################################################
 add_library(poly2tri STATIC
+
```
zeroset: Important new line? ;)

Issue #365, opened 2019-07-11
Waiting areas


chraibi:
In Gitlab by @anna-braun on Oct 4, 2018, 09:40 [[origin](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/issues/287)]

In order to simulate the transition from a platform edge to a train, waiting areas should be defined in front of the doors (transitions), for example, areas where only a certain number of people can stand.

chraibi:
In Gitlab by @chraibi on Oct 9, 2018, 22:01

assigned to @chraibi

chraibi:
In Gitlab by @chraibi on Oct 9, 2018, 22:01

assigned to @schroedtert and unassigned @chraibi

chraibi:
In Gitlab by @chraibi on Jan 28, 2019, 16:15

Any updates here?

chraibi:
In Gitlab by @chraibi on Mar 15, 2019, 16:12

Should this be closed?

chraibi:
In Gitlab by @schroedtert on Mar 21, 2019, 11:26

No it can not be closed. The distribution inside the waiting area needs to be adjusted be more realistic.

chraibi:
In Gitlab by @123morgoth123 on May 6, 2019, 19:12

I made some tests with a simple geometry to get to know waiting areas. Two issues occurred:
1. max_peds is ignored. In both cases (global_timer=”true” and “false”) more then the maximum amount of pedestrians are simultaneously in the waiting area
2. with global_timer=”false” two pedestrians are stuck in a waiting area

This folder contains the ini, geo and trajectory files: [(https://1drv.ms/f/s!AsV6LgpLPxYhgaMdAcUFsc6d6H5IPw) ]

chraibi:
In Gitlab by @chraibi on May 11, 2019, 00:39

Hi Tim, 
@123morgoth123 

thanks for your comment. 

This feature is still WIP. 

For the second case, it seems that agents 36 and 13 are heading in the wrong direction

![Screenshot_2019-05-11_at_00.37.33](https://gitlab.version.fz-juelich.de/JuPedSim/jpscore/uploads/f4fbca492d06fee19837f65e7cdafe57/Screenshot_2019-05-11_at_00.37.33.png)

![Screenshot_2019-05-11_at_00.37.47](/uploads/159c0cdb63d2c9a66c03c42c3d3421fa/Screenshot_2019-05-11_at_00.37.47.png)

[...]

PR #480, opened 2019-08-20
Update WaitingArea

schroedtert:
Remove some minor issues regarding the reading of waiting areas.
Addresses the issue of peds not getting distributed according to the given probabilities (see #365)

### DIFF
--- a/geometry/WaitingArea.h
+++ b/geometry/WaitingArea.h
@@ -157,8 +157,9 @@ public:
     void SetWaitingTime(int waitingTime);
 
     /**
-     * Returns the ID of the next goal based on the probability
-     * @return ID of the next goal
+     * Returns the ID of the next goal based on the probability. Only
+     * considers open goals.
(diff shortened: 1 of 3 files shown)


chraibi:
@zeroset can you please check and close if necessary?

geometry/WaitingArea.h
```diff
@@ -157,8 +157,9 @@ class WaitingArea : public Goal{
     void SetWaitingTime(int waitingTime);
 
     /**
-     * Returns the ID of the next goal based on the probability
-     * @return ID of the next goal
+     * Returns the ID of the next goal based on the probability. Only
+     * considers open goals.
+     * @return ID of the next goal or own ID if no open goal available
```
zeroset: Probably too much for this PR, but for functions returning "nothing" like no goal found we should use [std::optional](https://en.cppreference.com/w/cpp/utility/optional).

PR #483, opened 2019-08-26
First step to more general output

### LINKED ISSUE #418: More general output
Proposal:
Instead of only writing one of the properties to the output file, write all available features as velocity, router, intermediate/final goal, to output file. With this the wanted property can be displayed in jpsvis without need to run a simulation multiple times when different properties are needed.

schroedtert:
First additions to a more general output, speed and velocity are now written when using plain format (closes #418)


gjaeger:
looks [good](https://github.com/JuPedSim/jpscore/issues/418#issuecomment-524763169).

gjaeger:
Should we extend the element `<trajectories>` with `data_mode_detail`so that the output of the speed/velocity can be switched on or off (`true` or `false`)? 

Like
```
 <trajectories format="plain" fps="8" color_mode="velocity" data_mode_detail="true">
    <file location="trajectories.xml" />   
 </trajectories>
```

chraibi:
> Should we extend the element `<trajectories>` with `data_mode_detail`so that the output of the speed/velocity can be switched on or off (`true` or `false`)?
> 
> Like
> 
> ```
>  <trajectories format="plain" fps="8" color_mode="velocity" data_mode_detail="true">
>     <file location="trajectories.xml" />   
>  </trajectories>
> ```

Yes, I think this is important for large simulations, files tend to be bigger the more details you write out. 
for analysis only, you are only interested in the trajectories. Everything else is just unnecessary.

So, would you please enhance this with this option?

schroedtert:
I think I would even go further and use something like this:
```
<trajectories format="plain" fps="8" color_mode="velocity" data_mode_detail="true">
    <file location="trajectories.xml" />   
    <speed=“true”/>
    <velocity=“false”/>
    <final_target=“true”/>
    ....
 </trajectories>
```
When using `data_mode_detail="true”` the user can also select which attributes are wanted in the output. Using this, most importantly also **jpsvis** should be updated that each of the categories can be used as `color_mode`. This would make it easier to see different attributes without the need to restart the simulation.
What do you think?

gjaeger:
@schroedtert A very good suggestion. This can be further developed in the future.

chraibi:
> I think I would even go further and use something like this:
> 
> ```
> <trajectories format="plain" fps="8" color_mode="velocity" data_mode_detail="true">
>     <file location="trajectories.xml" />   
>     <speed=“true”/>
>     <velocity=“false”/>
>     <final_target=“true”/>
>     ....
>  </trajectories>
> ```
> 
> When using `data_mode_detail="true”` the user can also select which attributes are wanted in the output. Using this, most importantly also **jpsvis** should be updated that each of the categories can be used as `color_mode`. This would make it easier to see different attributes without the need to restart the simulation.
> What do you think?

Marvelous! 👍 

(I can update `jpsvis` when this is finished)

schroedtert:
Ok, then I need a list of potential outputs:
- speed
- velocity
- final goal
- current goal
- is waiting
Is there anything I forgot in this list?

gjaeger:
For velocity: v_x, v_y and v_z

schroedtert:
With the latest commit, optional output can be added in the following matter:
``` 
<trajectories format="plain"  fps="8">
      <file location="bottleneck_traj.xml" />
      <optional_output speed="FALSE" velocity="TRUE" final_goal="and"  intermediate_goal="false”/>
</trajectories>
``` 
At the moment only possible for speed, velocity (Vx and Vy as there is no Vz), final goal and intermediate goal. Additional options can be added easily and may come later. Is there anything important I forgot?

gjaeger:
It works. What about with "waiting"?

schroedtert:
> It works. What about with "waiting"?

There is some work which needs to be merged first.

chraibi:
What about the desired direction (e0)?

gjaeger:
> What about the desired direction (e0)?
@chraibi 
Do you mean `eO` (elipseOrientation)?

chraibi:
> > What about the desired direction (e0)?
> 
> @chraibi
> Do you mean `eO` (elipseOrientation)?

Not the ellipseOrientation, but the desired direction.

gjaeger:
> Not the ellipseOrientation, but the desired direction.

Thanks. I have added this to the list in issue [jpsreport/123](https://github.com/JuPedSim/jpsreport/issues/123#issuecomment-510629277).

schroedtert:
Will add it, when you tell me how to get the desired direction: Is it via `Pedestrian::GetLastE0()`?

schroedtert:
@chraibi I will add it, when you tell me how to get the desired direction: Is it via `Pedestrian::GetLastE0()`?

IO/IODispatcher.cpp
```diff
@@ -324,7 +321,86 @@ void TrajectoriesJPSV04::WriteFooter()
...
-TrajectoriesFLAT::TrajectoriesFLAT() : Trajectories() {}
+TrajectoriesFLAT::TrajectoriesFLAT() : Trajectories()
+{
+    // Add header, info and output for speed
+    _optionalOutputHeader[OptionalOutput::speed] = "V\t";
+    _optionalOutputInfo[OptionalOutput::speed]   = "#V: speed of the pedestrian (in m/s)";
+    _optionalOutput[OptionalOutput::speed]       = [](const Pedestrian * ped) {
+        std::stringstream stream;
+        stream << std::setprecision(2) << std::fixed << ped->GetV().Norm() << "\t";
+        return stream.str();
```
zeroset: Maybe you could use libfmt here as well? like: 
```
#include "general/Format.h"
...
return fmt::format(check:fmt("{:.2f}\t"), ped->GetV().Norm());
```

IO/IODispatcher.cpp
```diff
@@ -514,9 +589,20 @@ void TrajectoriesFLAT::WriteHeader(long nPeds, double fps, Building * building,
...
+    }
     Write("\n");
-    //Write("#ID\tFR\tX\tY\tZ");// @todo: maybe use two different formats
-    Write("#ID\tFR\tX\tY\tZ\tA\tB\tANGLE\tCOLOR"); // a b angle color
+
+    // Write("#ID\tFR\tX\tY\tZ");// @todo: maybe use two different formats
+    std::string header("#ID\tFR\tX\tY\tZ\tA\tB\tANGLE\tCOLOR\t");
+
+    // Add header for optional output options
+    for(auto option : _optionalOutputOptions) {
```
zeroset:
```suggestion
    for(const auto & option : _optionalOutputOptions) {
```

IO/IODispatcher.cpp
```diff
@@ -514,9 +589,20 @@ void TrajectoriesFLAT::WriteHeader(long nPeds, double fps, Building * building,
     Write("#ANGLE: orientation of the ellipse");
     Write("#COLOR: color of the ellipse");
 
+    // Add info for optional output options
+    for(auto option : _optionalOutputOptions) {
```
zeroset:
```suggestion
    for(const auto & option : _optionalOutputOptions) {
```

IO/IODispatcher.cpp
```diff
@@ -552,7 +641,13 @@ void TrajectoriesFLAT::WriteFrame(int frameNr, Building * building)
             b,
             phi * RAD2DEG,
             color);
-        Write(tmp);
+
+        std::string frame(tmp);
+        for(auto option : _optionalOutputOptions) {
```
zeroset:
```suggestion
        for(const auto & option : _optionalOutputOptions) {
```

IO/IODispatcher.cpp
```diff
@@ -591,7 +685,8 @@ void TrajectoriesVTK::WriteHeader(long nPeds, double fps, Building * building, i
 void TrajectoriesVTK::WriteGeometry(Building * building)
 {
     (void) building; // avoid warning
-    Log->Write("WARNING:\t Creating NavMesh is deprecated. Please have a look at old files on git "
+    Log->Write("WARNING:\t Creating NavMesh is deprecated. Please have a look at "
+               "old files on git "
```
zeroset: Those line breaks seem to be strange. Maybe put everything on a single line and let clang format split it again, hopefully just one time ...

IO/IniFileParser.cpp
```diff
@@ -365,9 +365,120 @@ bool IniFileParser::ParseHeader(TiXmlNode * xHeader)
                 _config->GetHostname().c_str(),
                 _config->GetPort());
         }
+
+        if(xTrajectories->FirstChild("optional_output")) {
+            Log->Write("WARNING: These optional options do only work with plain output format!");
```
zeroset: Please use the new logging mechanism

```suggestion
            Logging::Warning("These optional options do only work with plain output format!");
```
Don't forget to include `#include "general/Logger.h"`

IO/IODispatcher.cpp
schroedtert: Should I change it for every option?

IO/IODispatcher.cpp
schroedtert: Will be removed soon in an other PR soon. I just need to merge this PR and the one in the pipeline.

IO/IniFileParser.cpp
```diff
@@ -365,9 +366,120 @@ bool IniFileParser::ParseHeader(TiXmlNode * xHeader)
...
+                std::transform(in.begin(), in.end(), in.begin(), ::tolower);
+
+                if(in == "true") {
+                    _config->AddOptionalOutputOption(OptionalOutput::group);
+                    Log->Write("INFO: group added to output");
+                } else {
+                    Log->Write("INFO: group not added to output");
+                }
+            }
+        }
```
schroedtert: Sorry for still using the old logging, is changed in follow up PR (ready to review soon)

PR #486, opened 2019-09-04
Refactored ArgumentParser to use CLI11

Ozaq:
This also fixes a crashdump when calling jpscore without parameter and
no ini.xml in current working dir.

### DIFF
--- a/general/ArgumentParser.cpp
+++ b/general/ArgumentParser.cpp
@@ -1,8 +1,6 @@
 /**
  * \file        ArgumentParser.cpp
- * \date        Apr 20, 2009
- * \version     v0.8
- * \copyright   <2009-2018> Forschungszentrum Jülich GmbH. All rights reserved.
+ * \copyright   <2009-2019> Forschungszentrum Jülich GmbH. All rights reserved.
  *
  * \section License
  * This file is part of JuPedSim.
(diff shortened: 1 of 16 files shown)


general/ArgumentParser.cpp
```diff
@@ -34,123 +28,25 @@
...
+    try {
+        app.parse(argc, argv);
+        // Silence warnigns about iniFilePathOpt declared as member but never
+        // used. iniFilePathOpt is a member to keep the declaration of all
+        // options in the header file. If there was no iniFilePathOpt member the
+        // add_option(...) call would need to be done the constructor.
+        (void)iniFilePathOpt;
+        if (auto msg = CLI::ExistingFile(iniFilePath.string()); !msg.empty()) {
+            throw CLI::ValidationError(msg);
 }
```
zeroset: The format is broken.

main.cpp
```diff
@@ -54,14 +56,21 @@ int main(int argc, char** argv)
...
+    ArgumentParser a;
+    if(auto [execution, return_code] = a.Parse(argc, argv);
+       execution == ArgumentParser::Execution::ABORT) {
+        return return_code;
     }
+
+    Configuration config;
+
+    //TODO remove me in refactoring
+    Configuration* configuration = &config;
```
zeroset: Whats the problem with removing this pointer? Is it used somewhere outside this file?

general/ArgumentParser.cpp
Ozaq: > The format is broken.

? Sorry I don't get it

general/ArgumentParser.cpp
```diff
@@ -34,123 +28,25 @@
...
-          else if (!argument.compare(0, prefix1.size(), prefix1)) {
-               argument.erase(0, prefix1.size());
-          }
-          IniFileParser* p = new IniFileParser(_config);
-          bool status = p->Parse(argument);
-          delete p;
-          return status;
-     }
-
-#ifdef _JPS_AS_A_SERVICE //TODO try to avoid macros!
```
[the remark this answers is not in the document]
Ozaq: Hmm this is going to be, ehem... difficult. the removed section contains some additional flags that I cannot test since I cannot compile with _JPS_AS_A_SERVICE. I was under the impression this was defunct. If this is needed, I maybe should try to get it compiling again.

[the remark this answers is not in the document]
chraibi: `_JPS_AS_A_SERVICE` does not have to be tested. It sounds weird, but it's going to be used in an upcoming project. 
When this happens, we will have to deal with it in another PR.

PR #489, opened 2019-09-15
Build deps for windows

Ozaq:
Lets get a smooth setup for widows :)

- [ ] Write "How-To" Guide
- [ ] Add boost dependency to setup-deps 
- [ ] Add CGAL dependency to setup-deps
- [ ] Add convenience script for windows to "just build all deps"

### DIFF
--- a/scripts/bash.bat
+++ b/scripts/bash.bat
@@ -1,0 +1,5 @@
+@echo off
+%HOMEDRIVE%
+cd %HOMEPATH%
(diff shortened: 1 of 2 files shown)


chraibi:
This [wiki](//www.jupedsim.org/jpscore/2018-10-21-windows) may be relevant.
Here only VisualStudio is considered.

Ozaq:
After discussions with @zeroset it seems much more likely that building native on windows will be replaced by cross compiling for windows. Hence this is not needed atm.

PR #492, opened 2019-09-20
Fixed bug when reading schedule with header section in ini-file.

schroedtert:
Fixed a bug, where the schedule file was not read when inside the header section of the ini-file.


zeroset:
Shouldn't we "strictly" define where the schedule_file must be set?

schroedtert:
Problem is that we can either use a header section at the beginning of the ini-file or not. If I remember correctly the header section is the current standard but for compatibility reasons we still support ini-files without header section.

PR #495, opened 2019-09-23
Replaces crossing pointer with plain member variable

zeroset:
This Crossing belongs to a certain goal. No need for dynamic allocation
here.

### DIFF
--- a/geometry/Goal.cpp
+++ b/geometry/Goal.cpp
@@ -41,10 +41,6 @@ Goal::Goal()
      _subRoomID = -1;
 }
 
-Goal::~Goal()
-{
-}
-
 void Goal::AddWall(const Wall& w)
 {
      _walls.push_back(w);
@@ -240,8 +236,6 @@ bool Goal::ConvertLineToPoly()
           _crossing.SetPoint2(tmp_line.GetCentre());
      }
 
-
-//     std::cout << "Crossing goal: " << _crossing.GetUniqueID() << _crossing.toString() << std::endl;
      return true;
 }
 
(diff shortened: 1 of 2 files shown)


geometry/Goal.cpp
```diff
@@ -37,14 +37,12 @@ Goal::Goal()
...
      _walls = std::vector<Wall > ();
      _poly = std::vector<Point > ();
-     _crossing = new Crossing();
      _roomID = -1;
      _subRoomID = -1;
 }
 
 Goal::~Goal()
 {
-
```
schroedtert: What about using `~Goal() = default` in header?

geometry/Goal.cpp
```diff
@@ -234,16 +232,16 @@ bool Goal::ConvertLineToPoly()
...
      }else{
-          _crossing->SetPoint1(_poly[0]);
+          _crossing.SetPoint1(_poly[0]);
           Line tmp_line(_poly[_poly.size()/2], _poly[(_poly.size()/2)+1], 0);
-          _crossing->SetPoint2(tmp_line.GetCentre());
+          _crossing.SetPoint2(tmp_line.GetCentre());
      }
 
 
-//     std::cout << "Crossing goal: " << _crossing->GetUniqueID() << _crossing->toString() << std::endl;
```
schroedtert: Can be deleted. Blame me for the comment.

geometry/Goal.cpp
zeroset: Changed, thanks.

geometry/Goal.cpp
zeroset: removed ;)

PR #499, opened 2019-09-26
Adds SPDLog based logging

zeroset:
We need to exchange old logging with new SPDLog based Logger

### DIFF
--- a/IO/GeoFileParser.cpp
+++ b/IO/GeoFileParser.cpp
@@ -21,10 +21,12 @@
 #include "GeoFileParser.h"
 
 #include "general/Filesystem.h"
+#include "general/Logger.h"
 #include "geometry/SubRoom.h"
 #include "geometry/WaitingArea.h"
 #include "geometry/Wall.h"
 
+#include <fmt/format.h>
 #include <tinyxml.h>
 
 GeoFileParser::GeoFileParser(Configuration * configuration) : _configuration(configuration) {}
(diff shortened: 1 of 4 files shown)


IO/GeoFileParser.cpp
```diff
@@ -306,25 +311,25 @@ bool GeoFileParser::LoadGeometry(Building* building)
                p /= transFilename;
                transFilename = p.string();
 
-               Log->Write("INFO:\tParsing transition from file <%s>", transFilename.c_str());
+               Logging::Info(fmt::format("Parsing transition from file <{:s}>", transFilename));
```
chraibi: why fmt::format is used here?

zeroset: To insert the transFilename into the format string. Before this was done with sprintf

zeroset:
I had to remove some parts from @chraibi because they where wrong and i was unable to merge develop. Please let us merge the SPDLog integration step by step. This can be merged as soon as it is reviewed and tests are done.

main.cpp
```diff
@@ -53,11 +53,11 @@ int main(int argc, char ** argv)
 {
     Logging::Guard guard;
     Logging::Info("Starting JuPedSim - JPScore");
-    Logging::Info(fmt::format("Version {}", JPSCORE_VERSION));
-    Logging::Info(fmt::format("Commit id {}", GIT_COMMIT_HASH));
-    Logging::Info(fmt::format("Commit date {}", GIT_COMMIT_DATE));
-    Logging::Info(fmt::format("Build from branch {}", GIT_BRANCH));
-    Logging::Info(fmt::format("Build with {}({})", compiler_id, compiler_version));
+    Logging::Info(fmt::format(check_fmt("Version {}"), JPSCORE_VERSION));
```
chraibi: why do we have to use `check_fmt`?

zeroset: Thats the compile time check of the format string. Unfortunately we can not hide this inside a function.

PR #507, opened 2019-10-01
Adds clang-format-hooks as third party lib

zeroset:
See https://github.com/barisione/clang-format-hooks

This is the pre-commit hook for clang format. Everybody needs to activate this on their own since pre commit hooks can not be set "globally". Maybe we will add a pre-receive hook later on.  

Installing is done using the following command in the jpscore folder: 
```
./third-party/clang-format-hooks/git-pre-commit-format install
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
       dist: bionic
       env: TOOLCHAIN=clang
       script:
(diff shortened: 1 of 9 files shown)


zeroset:
I added a travis script to check the format using the apply format script from the added hooks. If the format script returns a diff, CI will fail. 
Additionally i added a clang format config in third-party folder to disable clang format on third party libraries.

PR #493, opened 2019-09-23
Refactoring router/direction strategies (#472)

schroedtert:
- Restructure in router and direction strategy
    * `routing` renamed in `router`: Now only contains the different routers
    * `routing/direction` renamed in `direction`:
        * moved already existing direction strategies to `direction/walking` and move them to separate files.
        * add `direction/waiting` for direction strategies related to waiting pedestrians
        * add `DirectionManager` for deciding if waiting or walking strategy should be applied, based on the state of the pedestrian
   * removed FF_trips after adding the functionality to normal FF router

* Add missing documentation.

* Bug-fixes to solve some segFaults related to WaitingAreas, GoalManager

* Update scenario_13_schedule and some clean up.

* Fixed bug when reading waiting areas. Update scenario_12_waiting_area and some clean up.

PR #514, opened 2019-10-16
Refactoring of GoalManager

schroedtert:
- GoalManager now works on building level and not only for waiting areas
- GoalManager controls the peds waiting status based on its position and the waiting area status
- Remove old Logger calls
- Fix segmentation fault when deleting Crossing which belongs to a Goal

First PR of divided #493 (merging waiting into develop). More to come.


IO/GeoFileParser.cpp
```diff
@@ -717,12 +717,13 @@ Goal * GeoFileParser::parseWaitingAreaNode(TiXmlElement * e)
         if(int value = xmltoi(attribute, -1); value > -1 && attribute == std::to_string(value)) {
             wa->SetRoomID(value);
         } else {
-            Log->Write("ERROR:\t  waiting area %d: room_id set but not an integer", wa->GetId());
+            Logging::Error(
+                fmt::format("waiting area {:d}: room_id set but not an integer", wa->GetId()));
```
zeroset: Sorry, but please add `check_fmt` Macro on format string to check the Format string at compile-time (see [documentation](https://fmt.dev/latest/index.html#safety)).
`check_fmt` itself is defined in `general/Format.h`. Please also use this include instead of including `<fmt::format.h>`.
```suggestion
                fmt::format(check_fmt("waiting area {:d}: room_id set but not an integer"), wa->GetId()));
```

geometry/Goal.cpp
```diff
@@ -225,8 +225,8 @@ bool Goal::ConvertLineToPoly()
         tmp    = _poly[_poly.size() / 2];
         diff   = point1 - tmp;
 
-        point1 = tmp + diff * 0.95;
-        point2 = tmp + diff * 0.05;
+        point1 = tmp + diff * 0.51;
```
zeroset: Why?

PR #521, opened 2019-10-27
Update logging

Ozaq:
EventManager.cpp now using new logging interface instead of old logging.

### DIFF
--- a/geometry/Crossing.cpp
+++ b/geometry/Crossing.cpp
@@ -27,6 +27,8 @@
 #include "Crossing.h"
 
 #include "SubRoom.h"
+#include "general/Format.h"
+#include "general/Logger.h"
 
 
 Crossing::Crossing()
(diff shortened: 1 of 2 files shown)


Ozaq:
@zeroset Since I had some time I have cleared another file. Sorry you will have to review again :D

geometry/Crossing.cpp
```diff
@@ -111,12 +113,12 @@ SubRoom * Crossing::GetOtherSubRoom(int roomID, int subroomID) const
     else if(_subRoom2->GetSubRoomID() == subroomID)
         return _subRoom1;
     else {
-        Log->Write(
-            "WARMING: \tCrossing::GetOtherSubRoom No exit found "
-            "on the other side\n ID=%hd, roomID=%hd, subroomID=%hd\n",
+        Logging::Warning(fmt::format(
+            check_fmt("Crossing::GetOtherSubroom No exit found on the other side ID={}, roomID={}, "
```
zeroset:
```suggestion
            check_fmt("Crossing::GetOtherSubRoom No exit found on the other side ID={}, roomID={}, "
```

geometry/Crossing.cpp
```diff
@@ -404,3 +387,30 @@ std::string Crossing::toString() const
     }
     return tmp.str();
 }
+
+namespace fmt
+{
+template <>
+struct formatter<Crossing> {
```
zeroset: :+1:

PR #522, opened 2019-10-28
Fix crash when no trajectories configured

Ozaq:
In case no trajectoy output file name was given jpscore tried to create
a folder with an empty path resulting in an unhandled exception.


Ozaq:
@zeroset I was updating the python tests when I encountered this issue.

PR #523, opened 2019-10-29
Exchange old Log with new SpdLog and libfmt based Logging

zeroset:


IO/IniFileParser.cpp
```diff
@@ -540,16 +545,17 @@ bool IniFileParser::ParseGCFMModel(TiXmlElement * xGCFM, TiXmlElement * xMainNod
...
-        _config->SetNuPed(atof(nu.c_str()));
-        _config->SetDistEffMaxPed(atof(disteff_max.c_str()));
-        _config->SetIntPWidthPed(atof(interpolation_width.c_str()));
-        Log->Write(
-            "INFO: \tfrep_ped nu=%.3f, dist_max=%.3f, disteff_max=%.3f, interpolation_width=%.3f",
-            atof(nu.c_str()),
-            atof(dist_max.c_str()),
-            atof(disteff_max.c_str()),
-            atof(interpolation_width.c_str()));
+        _config->SetMaxFPed(std::stod(dist_max));
```
Ozaq: This can now throw unlike the C conversion functions.

IO/IniFileParser.cpp
```diff
@@ -636,16 +643,17 @@ bool IniFileParser::ParseKrauszModel(TiXmlElement * xKrausz, TiXmlElement * xMai
...
```
Ozaq: This can now throw and should be handled.

IO/IniFileParser.cpp
```diff
@@ -656,16 +664,17 @@ bool IniFileParser::ParseKrauszModel(TiXmlElement * xKrausz, TiXmlElement * xMai
...
-            atof(disteff_max.c_str()),
-            atof(interpolation_width.c_str()));
+        _config->SetMaxFWall(std::stod(dist_max));
+        _config->SetNuWall(std::stod(nu));
+        _config->SetDistEffMaxWall(std::stod(disteff_max));
+        _config->SetIntPWidthWall(std::stod(interpolation_width));
+        Logging::Info(fmt::format(
+            check_fmt("Frep_wall mu={:.3f}, dist_max={:.3f}, disteff_max={:.3f}, "
+                      "interpolation_width={:.3f}"),
+            std::stod(nu),
```
Ozaq: This can now throw and should be handled.

geometry/Building.cpp
```diff
@@ -926,10 +925,10 @@ std::vector<Wall>
 Building::SplitWall(const std::shared_ptr<SubRoom> & subroom, const Wall & bigWall) const
 {
     std::vector<Wall> WallPieces;
-#if DEBUG
-    std::cout << subroom->GetSubRoomID() << "collect wall pieces with " << std::endl;
-    bigWall.WriteToErrorLog();
-#endif
+
```
Ozaq: If you refactor like this you probably should also add configurable log levels to the cmdline args. Currently the min log level to write out is "trace" i.e. all your debug messages  will end up in the log, always.

zeroset: Thanks for doing so in #526

main.cpp
```diff
@@ -67,10 +67,17 @@ int main(int argc, char ** argv)
     Configuration config;
     // TODO remove me in refactoring
     IniFileParser iniFileParser(&config);
-    if(!iniFileParser.Parse(a.IniFilePath())) {
+    try {
```
Ozaq: Ugh... Now you have two ways to indicate that the inifile could not be parsed, a return value of false and an exception.

Either, Or but not both please :) Id say `Parse()` should catch everything and return true/false.

zeroset: I will add Exception error-handling in IniFileParser in another PR.

Ozaq:
Looks good although I would recommend to cleanup the commits.

PR #531, opened 2019-11-04
Exchange old Log output with new Logging

zeroset:


geometry/SubRoom.cpp
```diff
@@ -591,9 +588,20 @@ bool SubRoom::Overlapp(const std::vector<Line *> & goals) const
             if(wall.Overlapp(*goal))
 
             {
-                wall.WriteToErrorLog();
-                std::cout << "Goal " << goal->GetPoint1()._x << ", " << goal->GetPoint1()._y << "| "
-                          << goal->GetPoint2()._x << ", " << goal->GetPoint2()._y << std::endl;
```
schroedtert: Error message is missing, just a wall and goal are printed

geometry/SubRoom.cpp
```diff
@@ -978,13 +989,6 @@ std::string NormalSubRoom::WritePolyLine() const
     return s;
 }
 
-void NormalSubRoom::WriteToErrorLog() const
-{
-    Log->Write("\t\tNormal SubRoom:\n");
-    for(auto && w : _walls) {
-        w.WriteToErrorLog();
-    }
-}
```
schroedtert: Why removed?

geometry/Transition.cpp
```diff
@@ -125,58 +127,16 @@ SubRoom * Transition::GetOtherSubRoom(int roomID, int subroomID) const
...
-            " <->\tRoom: %d [%s] SubRoom: %d\n",
-            GetRoom2()->GetID(),
-            GetRoom2()->GetCaption().c_str(),
-            GetSubRoom2()->GetSubRoomID());
-    } else {
-        sprintf(tmp, " <->\tAusgang\n");
-    }
-    s.append(tmp);
-    Log->Write(s);
-}
```
schroedtert: Why removed?

geometry/Wall.cpp
```diff
@@ -36,19 +36,6 @@ Wall::Wall(const Point & p1, const Point & p2, const std::string & type) : Line(
...
-    sprintf(
-        tmp,
-        "\t\tWALL: (%f, %f) -- (%f, %f)\n",
-        GetPoint1()._x,
-        GetPoint1()._y,
-        GetPoint2()._x,
-        GetPoint2()._y);
-    Log->Write(tmp);
-}
-
```
schroedtert: Why removed?

geometry/Room.cpp
```diff
@@ -161,20 +165,6 @@ void Room::AddSubRoom(SubRoom * r)
...
-    sprintf(tmp, "\tRaum: %d [%s]:\n", _id, _caption.c_str());
-    s.append(tmp);
-    Log->Write(s);
-    // SubRooms
-    for(int i = 0; i < GetNumberOfSubRooms(); i++) {
-        SubRoom * sub = GetSubRoom(i);
-        sub->WriteToErrorLog();
-    }
-}
-
```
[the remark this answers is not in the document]
zeroset: All functions like this were actually never used. We will have libfmt formatters later. So I think we don't need them anymore.

PR #534, opened 2019-11-07
Removes Gompertz Model

zeroset:
We are refactoring the simulation and stick to velocity and GCFM
Model for the moment. Gompertz Model will be reintroduced later.


xsd/jps_ini_core.xsd
```diff
@@ -428,7 +428,6 @@
                   <xs:attribute name="description" use="required">
                     <xs:simpleType>
                       <xs:restriction base="xs:string">
-                        <xs:enumeration value="gompertz" />
```
schroedtert: Isn’t the velocity model missing?

math/ForceModel.cpp
```diff
@@ -21,7 +21,7 @@
  * along with JuPedSim. If not, see <http://www.gnu.org/licenses/>.
  *
  * \section Description
- * Implementation of the GCFM and the Gompertz model.
+ * Implementation of the GCFM.
```
schroedtert: Is this class needed? It is only used in ODESolver, which itself is not used anywhere (as far as I can see)

zeroset: I think you are right. I am cleaning up the models step by step. Will be removed in a following pr.

xsd/jps_ini_core.xsd
zeroset: The whole xsd needs an update I think. Just wanted to remove gompertz...

Issue #482, opened 2019-08-22
Global toolset for random numbers


schroedtert:
**Is your feature request related to a problem? Please describe.**
At the moment we have multiple classes which need a random number. Since we don’t have a global interface to a random number engine, each class creates its own random number engine or in some parts still relies on `rand`.

**Describe the solution you'd like**
A global toolset for creating random numbers, best with interface to different distributions as described wanted in #318.

**Additional context**
I want to open this issue for discussion mostly about design such a toolset. As it is mostly an interface to different `std::random` calls I’m not sure if it is desirable. Ideas I had for designing:
- class with mostly static function calls
- namespace where the functions are provided
- class with singleton pattern, and then operating with the instance
Any suggestions and recommendations?

schroedtert:
See also #223

Issue #318, opened 2019-07-11
statistical distributions


chraibi:
In Gitlab by @gjaeger on Jun 24, 2019, 19:37 [[origin](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/issues/331)]

At the moment, the pre-movement time is Gauss-distributed. The agent parameters are also.

I suggest that we implement more statistical distributions (for agent velocity and pre-movement time):

- no distribution 
- uniform (min and max)
- truncated normal (min, max, mean, sigma)
- normal (mean, sigma) - already exists
- triangular (min, max, mean)
- log-normal (min, max, mean, sigma, sigma_2)

Issue #223, opened 2019-07-11
Cleanup the used random generators


chraibi:
The random engine should be initialized once (with seed).

prefer  `std::mersenne_twister_engine`.

zeroset:
See #482

PR #536, opened 2019-11-07
New random toolset

schroedtert:
As in #223 mentioned, we use a ton of different random number generators across jpscore. This is a first idea of a toolset to unify the RNG calls.

This is only a draft of the toolset, I want to discuss this first before I start to change the RNG calls around the project. If you have any comments on improvement, please let me know!

This is solves #482.
After finishing the refactoring #223 could be closed.
Some distributions mentioned in #318 are already included, if a std implementation of the distribution was available.


general/Random.h
```diff
@@ -0,0 +1,147 @@
...
+ * @return random number of a normal distribution with given parameters
+ */
+template <class RealType = double>
+RealType GetNormal(RealType mean, RealType stddev)
+{
+    if(!_init) {
+        throw std::logic_error("RandomNumber not yet seeded!");
+    }
+
+    std::normal_distribution dist(mean, stddev);
```
schroedtert: Is there a may to avoid creating a distribution in each function call?

general/Random.h
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * Returns a random number of a normal distribution with mean (\p mean) and standard deviation (\p stddev).
+ * @tparam RealType
+ * @param mean mean of the normal distribution
+ * @param stddev standard deviation of the normal distribution
+ * @return random number of a normal distribution with given parameters
+ */
+template <class RealType = double>
```
zeroset: I would prefer `typename` instead of `class`. Especially since RealType is not a class in most cases.

general/Random.h
```diff
@@ -0,0 +1,147 @@
...
+ * @tparam RealType
+ * @param mean mean of the normal distribution
+ * @param stddev standard deviation of the normal distribution
+ * @return random number of a normal distribution with given parameters
+ */
+template <typename RealType = double>
+RealType GetLogNormal(RealType mean, RealType stddev)
+{
+    if(!_init) {
+        throw std::logic_error("RandomNumber not yet seeded!");
```
zeroset: Maybe we should use `assert` here. If you use an uninitialized random number generator it's more like a programming error ...

Ozaq:
Instead of relying on a global random number generator i would much more like to see something like a `SimulationContext` that owns the contextual state surrounding the Simulation (e.g. rng)

Also I think the implementation should be thread-safe (I did not check where currently random numbers are requested, it might be that it is not necessary but I very much suspect that it is)

PR #546, opened 2019-11-10
Refactor project layout

Ozaq:
libcore and jpscore now reside in a seperate directory. This should
allow us to integrate jpsreport easier.


Ozaq:
@zeroset should work now, can you review and rebase merge it?

Issue #553, opened 2019-11-14
clean_geometry test 2 hanging in Building::correct on MacOS


zeroset:
Sometimes the geometry test case 2 is hanging on MacOS when correcting the geometry. 
The following Loop is never left again: 
```
for(auto wallPiece : WallPieces) {
    std::vector<Wall> tmpWallPieces;
    tmpWallPieces = SplitWall(subroom.second, wallPiece);
    if(!tmpWallPieces.empty()) {
        // std::cout << "set ok because tmp size =" << tmpWallPieces.size() << std::endl;
        ok = 0; /// stay in the loop
        // append tmpWallPieces to WallPiece
        WallPieces.insert(
            std::end(WallPieces),
            std::begin(tmpWallPieces),
            std::end(tmpWallPieces));
        // remove the line since it was split already
        auto it = std::find(WallPieces.begin(), WallPieces.end(), wallPiece);
        if(it != WallPieces.end()) {
            // std::cout<< KGRN << "delete wall ..." << RESET <<std::endl;
            // wallPiece.WriteToErrorLog();
            WallPieces.erase(it);
        }

        // std::cout << "BNow Wall peces size : " <<  WallPieces.size() << std::endl;
    }
}
```

It looks like one wall has a point with x coordinate almost +INF because it used to be something around `1.0 e154`

zeroset:
@chraibi @schroedtert can you try to reproduce by starting jpscore with `systemtest/test_geometry/clean_geometry/inifile2.xml`

schroedtert:
approx 1/5 cases it get stuck here

Ozaq:
Not related to MacOS :)

I could reproduce it under ASAN:
```
==69035==ERROR: AddressSanitizer: heap-use-after-free on address 0x613000000ea8 at pc 0x000111495e12 bp 0x7ffee0f7c150 sp 0x7ffee0f7b900
READ of size 60 at 0x613000000ea8 thread T0
    #0 0x111495e11 in __asan_memcpy (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x5ee11)
    #1 0x10fa280a8 in Line::Line(Line const&) Line.h:42
    #2 0x10fa28396 in Wall::Wall(Wall const&) Wall.h:31
    #3 0x10fa0c60c in Wall::Wall(Wall const&) Wall.h:31
    #4 0x10f01dd32 in Building::correct() const Building.cpp:772
[...]

PR #555, opened 2019-11-14
Band-aid-fix for use-after-free

Ozaq:
This commit is not a proper fix but should address the use after free in
most cases. The cause for the use-after-free was a modification
(push_back) of a vector while iterating over the same vector with a
range based for loop. The moment the vector has to be resized the
iterators of the loop get invalidated and we work on freed memory.

The "hack" is now to preemtively reserve the vector to a size that is
very unlikely to reach. The code has to be rewritten properly but for
now this should be good enough to buy us some time to do so.

Fixes: #553

### DIFF
--- a/libcore/src/geometry/Building.cpp
+++ b/libcore/src/geometry/Building.cpp
@@ -756,8 +756,16 @@ bool Building::correct() const
                 // bigWall.WriteToErrorLog();
                 //special treatment for doors
                 /////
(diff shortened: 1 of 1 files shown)


Ozaq:
@zeroset please have a look and decide if you want this band-aid-fix. I am not a fan of this "fix" but id much rather see JPScore not crash / hang for now until we can cleanup the actual issue.

PR #557, opened 2019-11-15
Change approch to testing

Ozaq:
Now tests are always build and instrumented with asan. A set of system
tests now also run with asan enabled.

No asan builds / tests added for windows since it is not available for
msvc at this time.

NOTE: this pr needs #553 to be fixed first (tests now show the issue and thus are failing)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,12 +39,6 @@ check_prefix_path()
 ################################################################################
 # Optional features
 ################################################################################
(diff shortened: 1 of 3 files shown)


zeroset:
I would prefer keeping the `BUILD_TESTING` var to enable testing. I think there are still ppl building this from source for use without tests. They don't need tests and asan. At least it would be nice to keep this option for people?

PR #563, opened 2019-11-22
Fixed the cmakelists changes the directory of systemtests

zeroset:
The system test directory was not adjusted in cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,7 @@ if (BUILD_TESTING)
         )
     endforeach()
 
-    file(GLOB_RECURSE test_py_files "${CMAKE_SOURCE_DIR}/Utest/*runtest_*.py")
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -286,7 +286,7 @@ if (BUILD_TESTING)
         )
     endforeach()
 
-    file(GLOB_RECURSE test_py_files "${CMAKE_SOURCE_DIR}/Utest/*runtest_*.py")
+    file(GLOB_RECURSE test_py_files "${CMAKE_SOURCE_DIR}/systemtest/*runtest_*.py")
```
schroedtert: Is there a file which matches `*runtest_*.py`?

zeroset: Yes ;) for example: `/systemtest/juelich_tests/test_1/runtest_juelich_1.py`

Issue #374, opened 2019-07-11
Improve FF


chraibi:
In Gitlab by @schroedtert on Apr 12, 2019, 15:16 [[origin](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/issues/311)]

* [x] Merge FF and FF_Trips done with 6789d7869d764f4f5b460f36277174dbf65b9e25
* [x] Remove `ERROR: 	Could not find a route for pedestrian 37 in room 2 and subroom 0` when ped leaves the geometry through door
* [x] If doors are temp_close at start of simulation do not quit computation
* [ ]  ~~Extend FF such that it is also possible to navigate to points not only exits~~ (does not belong here, needs to be added as feature in DirectionManager)

chraibi:
In Gitlab by @schroedtert on Apr 12, 2019, 15:24

changed the description

chraibi:
In Gitlab by @schroedtert on Apr 12, 2019, 15:31

created branch [`311-improve-ff`](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/compare/develop...311-improve-ff) to address this issue

chraibi:
In Gitlab by @schroedtert on Jun 27, 2019, 08:10

changed the description

PR #565, opened 2019-11-24
Refactoring FF Router

### LINKED ISSUE #564: traffic_constraints ignored
When testing the schedule function I wanted to "temp_close" all doors with by traffic_constraints. But this does not work, the traffic_constraints were ignored. Using an event file to "temp_close" the doors works. But it's a bug anyway.

 [Archiv.zip](https://github.com/JuPedSim/jpscore/files/3879130/Archiv.zip)

schroedtert:
- remove wrong error message from router, when ped is already outside
- remove not needed code
- adding missing documentation
- fix clang-tidy warnings (some)
- at start up compute all floor fields even if door is closed, penalise closed doors with increases edge weight

Addresses some of the issue mentioned in #374.

Fixes #564.
Needs some further testing, but maybe you can already have a first look.


anna-braun:
Yesterday I tested a funktion for issue #564 with this PR and it works fine. Today I wanted to test something for sources, still using this PR, but the simulation failed. In develop it works fine, so there seems to be a problem in this PR.
[Archiv.zip](https://github.com/JuPedSim/jpscore/files/3950540/Archiv.zip)

PR #591, opened 2019-12-04
Add missing check for global timer for WA

### LINKED ISSUE #590: Waiting_areas - Error control
When using global_timer="true" min_peds is not needed. But if min_peds is not used, the simulation crashes because of the missing value.

schroedtert:
Adds extra check, if global timer for waiting areas is set, if so min_ped do not need to be set.

Fixes #590


libcore/src/IO/GeoFileParser.cpp
```diff
@@ -837,10 +837,11 @@ Goal * GeoFileParser::parseWaitingAreaNode(TiXmlElement * e)
     // Additional checks:
     const bool waitPed =
         (wa->GetMinNumPed() > 0 && wa->GetMaxNumPed() > 0 && wa->GetWaitingTime() >= 0.);
-    const bool waitDoor = (wa->GetTransitionID() > 0);
+    const bool waitDoor   = (wa->GetTransitionID() > 0);
+    const bool waitGlobal = (wa->IsGlobalTimer() && wa->GetWaitingTime() >= 0.);
 
     // Either (minPed, maxPed, waitingTime) OR transitionID are set
-    if(!waitPed && !waitDoor) {
+    if(!waitPed && !waitDoor && !waitGlobal) {
```
Ozaq: The comment above does not fit the code. It should probably just be removed.

libcore/src/IO/GeoFileParser.cpp
```diff
@@ -837,10 +837,11 @@ Goal * GeoFileParser::parseWaitingAreaNode(TiXmlElement * e)
...
         Logging::Error(fmt::format(
             check_fmt("waiting area {:d}: min_peds, max_peds, waiting_time, transition_id not set "
                       "properly. Set either (min_peds, max_peds, waiting_time) or transition_id."),
```
Ozaq: The error message needs to be updated to reflect the code change.

libcore/src/IO/GeoFileParser.cpp
```diff
@@ -837,10 +837,11 @@ Goal * GeoFileParser::parseWaitingAreaNode(TiXmlElement * e)
     // Additional checks:
     const bool waitPed =
         (wa->GetMinNumPed() > 0 && wa->GetMaxNumPed() > 0 && wa->GetWaitingTime() >= 0.);
-    const bool waitDoor = (wa->GetTransitionID() > 0);
+    const bool waitDoor   = (wa->GetTransitionID() > 0);
+    const bool waitGlobal = (wa->IsGlobalTimer() && wa->GetWaitingTime() >= 0.);
```
Ozaq: You could rename the constants to improve readability quite a bit, e.g.:
```cpp
const bool hasValidConstraints = ...
const bool hasValidTransitionId = ...
const bool usesGlobalTimer = ...
if (!hasValidConstraints && !hasValidTransitionId && !usesGlobalTimer)
{
// handle error
}
```

libcore/src/IO/GeoFileParser.cpp
schroedtert: comment removed

libcore/src/IO/GeoFileParser.cpp
schroedtert: done

libcore/src/IO/GeoFileParser.cpp
schroedtert: done

Issue #572, opened 2019-12-03
Report: add commit hash to log output. 


zeroset:
Add the commit hash to log output for debugging purposes.
Should look as follows: 
```
% jpsreport_debug --version
----
JuPedSim - JPSreport

Current date   : Sun Nov 17 11:19:59 2019
Compiler        : clang++ (11.0.0)
Python           : /usr/local/bin/python3 (3.7.4)
Commit Hash : ecee019d
Commit Date  : Sun Nov 17 09:19:59 2019
----
```

Issue #574, opened 2019-09-06
Clean up cmake file 


chraibi:
`jpscore`'s build process got some lifting and [cleanup](https://github.com/JuPedSim/jpscore/pull/440). 

`jpsreport` needs the same treatment as well.

Ozaq:
@chraibi & @zeroset & @gjaeger I think this can be closed after integrating report into jpscore

Ozaq:
err.... I had a look into the file this DEFINITELY needs cleanup :P. Ill open a PR :/

PR #604, opened 2019-12-14
Reintroduce git info to jpsreport

Ozaq:
jpsreport now again properly reports git commit and change when running.
This was a left-over from the integration.

Fixes: #572

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,7 +224,8 @@ else()
     set(GIT_BRANCH "UNKNONW")
 endif()
 
-list(APPEND GIT_BUILD_INFO_DEFINES
+add_library(git-info INTERFACE)
+target_compile_definitions(git-info INTERFACE
     GIT_COMMIT_HASH="${GIT_SHA1}"
     GIT_COMMIT_DATE="${GIT_DATE}"
     GIT_COMMIT_SUBJECT="${GIT_COMMIT_SUBJECT}"
--- a/jpsreport/CMakeLists.txt
+++ b/jpsreport/CMakeLists.txt
@@ -319,11 +319,13 @@ target_link_libraries(report PUBLIC
     tinyxml
     fs
     Boost::boost
+    git-info
 )
 
 target_compile_definitions(report PUBLIC
     PYTHON="${PYTHON_EXECUTABLE}"
     PYTHON_VERSION="${PYTHON_VERSION_STRING}"
+    JPSREPORT_VERSION="${PROJECT_VERSION}"
 )
 
 add_executable(jpsreport
--- a/jpsreport/general/ArgumentParser.cpp
+++ b/jpsreport/general/ArgumentParser.cpp
@@ -102,11 +102,11 @@ void Logs()
     // first logs will go to stdout
     Log->Write("----\nJuPedSim - JPSreport\n");
     Log->Write("Current date   : %s", currentTime.c_str());
-    // Log->Write("Version        : %s", JPSREPORT_VERSION);
+    Log->Write("Version        : %s", JPSREPORT_VERSION);
     Log->Write("Compiler       : %s (%s)", true_cxx.c_str(), true_cxx_ver.c_str());
-    // Log->Write("Commit hash    : %s", GIT_COMMIT_HASH);
-    // Log->Write("Commit date    : %s", GIT_COMMIT_DATE);
-    // Log->Write("Branch         : %s", GIT_BRANCH);
+    Log->Write("Commit hash    : %s", GIT_COMMIT_HASH);
+    Log->Write("Commit date    : %s", GIT_COMMIT_DATE);
+    Log->Write("Branch         : %s", GIT_BRANCH);
     Log->Write("Python         : %s (%s)\n----\n", PYTHON, PYTHON_VERSION);
 }
 
--- a/libcore/CMakeLists.txt
+++ b/libcore/CMakeLists.txt
@@ -199,7 +199,6 @@ target_compile_options(core PRIVATE
 target_compile_definitions(core PUBLIC
     $<$<BOOL:${JPSFIRE}>:JPSFIRE>
     JPSCORE_VERSION="${PROJECT_VERSION}"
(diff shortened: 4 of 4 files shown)


gjaeger:
It seems to be working. 

```zsh
$ jpsreport --version

----
JuPedSim - JPSreport

Current date   : Sat Dec 14 11:20:17 2019
Version        : 0.8.5
Compiler       : clang++ (11.0.0)
Commit hash    : v0.8.4-881-gb6522ee5
Commit date    : Sat Dec 14 10:56:33 2019
Branch         : pr604
Python         : /usr/local/bin/python3 (3.7.5)
----
```

jpsreport/general/ArgumentParser.cpp
```diff
@@ -102,11 +102,11 @@ void Logs()
...
-    // Log->Write("Version        : %s", JPSREPORT_VERSION);
+    Log->Write("Version        : %s", JPSREPORT_VERSION);
     Log->Write("Compiler       : %s (%s)", true_cxx.c_str(), true_cxx_ver.c_str());
-    // Log->Write("Commit hash    : %s", GIT_COMMIT_HASH);
-    // Log->Write("Commit date    : %s", GIT_COMMIT_DATE);
-    // Log->Write("Branch         : %s", GIT_BRANCH);
+    Log->Write("Commit hash    : %s", GIT_COMMIT_HASH);
+    Log->Write("Commit date    : %s", GIT_COMMIT_DATE);
+    Log->Write("Branch         : %s", GIT_BRANCH);
     Log->Write("Python         : %s (%s)\n----\n", PYTHON, PYTHON_VERSION);
```
gjaeger: This output is no longer required. We removed the option to plot timeseries and Voronoi diagrams with python.

jpsreport/general/ArgumentParser.cpp
```diff
@@ -102,11 +102,11 @@ void Logs()
     // first logs will go to stdout
     Log->Write("----\nJuPedSim - JPSreport\n");
     Log->Write("Current date   : %s", currentTime.c_str());
-    // Log->Write("Version        : %s", JPSREPORT_VERSION);
+    Log->Write("Version        : %s", JPSREPORT_VERSION);
     Log->Write("Compiler       : %s (%s)", true_cxx.c_str(), true_cxx_ver.c_str());
-    // Log->Write("Commit hash    : %s", GIT_COMMIT_HASH);
-    // Log->Write("Commit date    : %s", GIT_COMMIT_DATE);
-    // Log->Write("Branch         : %s", GIT_BRANCH);
+    Log->Write("Commit hash    : %s", GIT_COMMIT_HASH);
```
gjaeger: Is there a _tag_ mistake?
`Commit hash    : v0.8.4-881-gb6522ee5`

jpsreport/CMakeLists.txt
```diff
@@ -319,11 +319,13 @@ target_link_libraries(report PUBLIC
     tinyxml
     fs
     Boost::boost
+    git-info
 )
 
 target_compile_definitions(report PUBLIC
     PYTHON="${PYTHON_EXECUTABLE}"
     PYTHON_VERSION="${PYTHON_VERSION_STRING}"
+    JPSREPORT_VERSION="0.8.5"
```
gjaeger: Why do you set the version here manually?

CMakeLists.txt
```diff
@@ -224,7 +224,8 @@ else()
     set(GIT_BRANCH "UNKNONW")
 endif()
 
-list(APPEND GIT_BUILD_INFO_DEFINES
+add_library(git-info INTERFACE)
+target_compile_definitions(git-info INTERFACE
```
gjaeger: Do jpscore and jpsreport use these lines together?

jpsreport/general/ArgumentParser.cpp
Ozaq: I can remove it but it does not really have anything to do with this PR since I didn't change the line

jpsreport/CMakeLists.txt
Ozaq: Ah right, I wanted to ask you or @zeroset if both report and core should use the same version number but then I forgot. So should they share the version number?

jpsreport/general/ArgumentParser.cpp
Ozaq: It is the same pattern as used in jps-report
Read it as:
{last tag}-{commits since last tag}-{short commit id}

CMakeLists.txt
Ozaq: So it works like this:
in line 227 a new "interface" library is defined that allows to propagate compilation flags, include paths, preprocessor definitions etc. by linking to this library
So jpsreport and jpscore depend on this library (see the respective target_link_libraries calls) and the defines will be set during the compilation of the source files.
TL;DR; yes :)

jpsreport/general/ArgumentParser.cpp
gjaeger: Okay. I remember it for a future pull request.

jpsreport/general/ArgumentParser.cpp
gjaeger: ok. Thanks.

jpsreport/CMakeLists.txt
gjaeger: A good question for @zeroset and @chraibi.

CMakeLists.txt
gjaeger: Thanks.

jpsreport/CMakeLists.txt
Ozaq: I talked to @zeroset in chat and the conclusion was that core and report will use the same verison number from now on

gjaeger: ok

jpsreport/general/ArgumentParser.cpp
Ozaq: @gjaeger I will fix this when cleaning up the cmakelists. See #574

PR #610, opened 2019-12-15
Refactoring log

Ozaq:
Easier logging with log macro

### DIFF
--- a/jpscore/src/main.cpp
+++ b/jpscore/src/main.cpp
@@ -30,7 +30,6 @@
 #include "general/ArgumentParser.h"
 #include "general/Compiler.h"
 #include "general/Configuration.h"
-#include "general/Format.h"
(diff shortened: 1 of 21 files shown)


Ozaq:
After refactoring logging to use spd log and libfmt the log messages got
rather unwieldy and inconvenient due to the compile time format string
check macro FMT_STRING which cannot be extracted into a function.

Due to this I introduced a set of log macros to make the code more
readable and less horrible to write:

LOG_(DEBUG|INFO|WARNING|ERROR) are now available. Please do no longer
use e.g. Logging::Info directly.

Be aware that due to limitations in the compile time format string check
the macros do not work with dynamic strings as format string.

Use LOG_ERROR("{}", myDynamicString) to log a message from a dynamic
string.

PR #611, opened 2019-12-15
Refactoring FF Router

### LINKED ISSUE #564: traffic_constraints ignored

schroedtert:
- remove wrong error message from router, when ped is already outside
- remove not needed code
- adding missing documentation
- fix clang-tidy warnings (some)
- at start up compute all floor fields even if door is closed, penalise closed doors with increased edge weight

Addresses some of the issue mentioned in #374.

Fixes #564.


libcore/src/Simulation.cpp
```diff
@@ -313,19 +325,21 @@ void Simulation::UpdateRoutesAndLocations()
...
-        int subRoomID  = ped->GetSubRoomID();
-
-        if(auto cross = dynamic_cast<Crossing *>(target)) {
-            if(cross->IsInRoom(roomID) && cross->IsInSubRoom(subRoomID)) {
-                if(!ped->IsWaiting() && cross->IsTempClose()) {
-                    ped->StartWaiting();
-                }
+        int goal = ped->FindRoute();
+        if(goal != FINAL_DEST_OUT) {
+            Hline * target = _building->GetTransOrCrossByUID(goal);
```
zeroset: const?

libcore/src/Simulation.cpp
```diff
@@ -313,19 +325,21 @@ void Simulation::UpdateRoutesAndLocations()
...
+            int roomID     = ped->GetRoomID();
+            int subRoomID  = ped->GetSubRoomID();
+
+            if(auto cross = dynamic_cast<Crossing *>(target)) {
```
zeroset: const

libcore/src/Simulation.cpp
schroedtert: done

libcore/src/Simulation.cpp
schroedtert: done

Issue #622, opened 2019-12-19
FF Router does not work with subrooms correctly


schroedtert:
When using a room with subrooms it may happen that one/mutliple of the subrooms do not get computed at all. The error arises as one point of the subroom is needed to assign the grid points of a subroom to the corresponding subroom. For computation of this point a door of the room is considered, as this door does not necessarily lies in all subrooms, the grid points do not get assigned correctly and the computation of the FFs breaks.
 
AC
- [x] fix bug (with f794ca4)
- [x] check if grid points need to be assigned to subroom (are needed)
- [x] check if the random point is really needed to assign point to a subroom (seems to be fastest solution at the moment)

schroedtert:
Closed with f794ca4

PR #623, opened 2019-12-19
Bug fix FF Router, when using subrooms

schroedtert:
- Fixes bug, that not all subrooms were computed correctly

A point inside a subroom is needed to assign each point inside the subroom to the subroom. Before the point was taken from the first door in doors array, which contains doors from the whole room. Hence it was possible that the door is not part of the subroom, which lead to errors when creating the FF of the subroom. Now the sample point is taken from one of the doors of the subroom.

See #622


zeroset:
Could you write a unittest for the exact bug?

schroedtert:
@zeroset tried to write a test. Any comments and remarks are highly appreciated!

schroedtert:
Okay, something is broken with the demo of the FF Router ... I will have a look where the problem is.

Issue #505, opened 2019-09-27
Remove commented code


zeroset:
As realized in #504 and #500 we have big chunks of commented code which is of no use anymore. In case of doubt we should use git to revert to old implementations. reformatting would mess up those comment blocks.

AC: 
- remove all code blocks commented out.

PR #624, opened 2019-12-20
Remove commented code

Ozaq:
There was a lot of commented code that made some files very noisy.

Commented code now removed and in some places superflous coments have
been removed as well.

Fixes: #505

### DIFF
--- a/jpscore/src/main.cpp
+++ b/jpscore/src/main.cpp
@@ -71,41 +71,28 @@ int main(int argc, char ** argv)
         return EXIT_FAILURE;
     }
 
-
-    // create and initialize the simulation engine
-    // Simulation
     time_t starttime, endtime;
     time(&starttime);
 
     Simulation sim(&config);
 
     if(sim.InitArgs()) {
-        // evacuation time
         double evacTime = 0;
         LOG_INFO("Simulation started with {} pedestrians", sim.GetPedsNumber());
         if(sim.GetAgentSrcManager().GetMaxAgentNumber()) {
-            // Start the thread for managing the sources of agents if any
-            // std::thread t1(sim.GetAgentSrcManager());
             double simMaxTime = config.GetTmax();
             std::thread t1(&AgentsSourcesManager::Run, &sim.GetAgentSrcManager());
-            while(!sim.GetAgentSrcManager().IsRunning()) {
-                // std::cout << "waiting...\n";
-            }
-            // main thread for the simulation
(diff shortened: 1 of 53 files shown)


jpscore/src/main.cpp
```diff
@@ -74,41 +73,28 @@ int main(int argc, char ** argv)
...
     if(sim.InitArgs()) {
-        // evacuation time
         double evacTime = 0;
         LOG_INFO("Simulation started with {} pedestrians", sim.GetPedsNumber());
         if(sim.GetAgentSrcManager().GetMaxAgentNumber()) {
-            // Start the thread for managing the sources of agents if any
-            // std::thread t1(sim.GetAgentSrcManager());
             double simMaxTime = config.GetTmax();
             std::thread t1(&AgentsSourcesManager::Run, &sim.GetAgentSrcManager());
-            while(!sim.GetAgentSrcManager().IsRunning()) {
```
chraibi: Any reason why you removed this while?

Ozaq: Hi @chraibi good catch /  question! :1st_place_medal: 

I did remove it because "Hey the loop is not doing anything expect spinning w.o. purpose, i.e. no loop body". 

However:
Your review remark promoted me to look into the code and be sure there are no side effects: Turns out there are! :/ The loop is actually blocking the main thread until `AgentSrcManager` has entered its `Run` method. Now the thing is: The loop is probably still useless because although it blocks the main thread until Run has been entered no additional work (that I can see) has been completed before continuing in the main loop. If the intention was here to sync execution of the `AgentSrcManager` with the main sim loop then it is questionable if this works as intended.

In conclusion, I would still remove this loop.

chraibi: Maybe this leads us to another issue, to be opened soon, about "How to force both threads to run nearly at sync?"

Ozaq: Actually Id much prefer to just remove the thread completely: See #628

libcore/src/Simulation.h
```diff
@@ -198,13 +185,9 @@ class Simulation
     bool correctGeometry(std::shared_ptr<Building> building, std::shared_ptr<TrainTimeTable>);
 
     /**
-     * Updates the output filename if the current file exceeds 10MB.
+     * Updates the output filename if the current file exceeds 16MB.
```
zeroset: I would prefer mentioning `_maxFileSize` here, instead of 16 MB.

Ozaq: Fixed

Issue #488, opened 2019-09-15
Exchange calls to old Logging with new SPD Logger


zeroset:
After the merge of #479 we need to exchange all calls to the old logger to the new one. Therefore the calls like: 
```c++
Log->Write("INFO: \tStart runSimulation() with %d pedestrians", sim.GetPedsNumber());
```
Need to be changed to:
```c++
Logging::Info(fmt::format("Simulation started with {} pedestrians", sim.GetPedsNumber()));
```
Additionally the old logging includes need to be removed and the new ones added. 
As an example you can look in to the files changed in #479 `main.cpp`.

For formatting libfmt should be used. Find the documentation about libfmt [here](https://fmt.dev/latest/).
Please also use the corresponding logging level (e.g. info, warning, error). The available logging functions can be found in [general/Logger.h](https://github.com/JuPedSim/jpscore/blob/e66f3149aae69216b58b556463d6136e42d0e1bf/general/Logger.h#L10)

see #499

PR #625, opened 2019-12-21
Remove old style logging

Ozaq:
HoHoHo! Christmas is a earlyt this year!
```
    ~*~
    /@\
   /%;@\
  o/@,%\o
  /%;`@,\
 '^^^H^^^`
```
Old style logging had several issues and hence has been replaced by
spdlog. New-style logging is already part of the codebase and this
commit removes the remaining old style log calls. Further the log option
is now also gone from the config as logging is now configured ad-hoc
from the command line only.

Fixes: #488

### DIFF
--- a/libcore/src/math/VelocityModel.cpp
+++ b/libcore/src/math/VelocityModel.cpp
@@ -29,6 +29,7 @@
 #include "VelocityModel.h"
 
 #include "direction/walking/DirectionStrategy.h"
+#include "general/Logger.h"
 #include "general/OpenMP.h"
 #include "geometry/SubRoom.h"
 #include "geometry/Wall.h"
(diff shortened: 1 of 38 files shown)


libcore/src/math/VelocityModel.cpp
```diff
@@ -75,14 +76,14 @@ bool VelocityModel::Init(Building * building)
...
-                "ERROR:\tVelocityModel::Init() cannot initialise route. ped %d is deleted in Room "
+            LOG_ERROR(
+                "VelocityModel::Init() cannot initialise route. ped {:d} is deleted in Room "
                 "%d %d.\n",
                 ped->GetID(),
                 ped->GetRoomID(),
                 ped->GetSubRoomID());
             building->DeletePedestrian(ped);
-            Log->incrementDeletedAgents();
+            // TODO KKZ track deleted peds
```
chraibi: :+1

Ozaq: Ah i should have written about that in the PR description: I think we should add a `stats` object to keep track of things like this but I wanted to keep this out if this PR

Ozaq: Now tracked in  #627

PR #631, opened 2019-12-23
Fix 574 - Cleanup CMakeLists for jpsreport

Ozaq:
Mostly removed commented out code. 
Also propagated the common compile options to libreport and jpsreport.

@gjaeger can you please look at the change in `Method_I.cpp`, I got a compiler warning there about an unused parameter in a setter. Due to this the setter was always setting the flag to true and never false.


jpsreport/methods/Method_I.cpp
```diff
@@ -294,7 +294,7 @@ void Method_I::GetIndividualFD(
 
 void Method_I::SetCalculateIndividualFD(bool individualFD)
 {
-    _calcIndividualFD = true;
+    _calcIndividualFD = individualFD;
```
gjaeger: I haven't had any warning yet. I had chosen this setting because I need the output IndividualFD for method I. The true/false option is not required.

We can remove this unnecessary code in another PR.

Ozaq: Alright! I removed the commit.

PR #640, opened 2020-01-05
JPSFIRE no longer optional

Ozaq:
Discussed this with Mohcine: JPSFIRE is now integral part of jpscore.
This should reduce build / test complexity.


Ozaq:
> Could you change the following pages too:
> http://www.jupedsim.org/jupedsim_install_on_linux.html#cmake-configuration-flags
> (docs/pages/jupedsim/)

Err... there is no mention of `JPSFIRE` on the page, nor anywhere else in the code

zeroset:
I have to install zlib on the Windows Build Agent, after that, the tests should work.

PR #642, opened 2020-01-07
Replace raw pointer array with map for neighborhood computation.

schroedtert:
Instead of using the raw pointer array a map is used for the local copy of the pedestrians for the neighborhood computation. Maybe solving the issue with incredible large memory usuage.

@Ozaq could you send me the simulation with the large memory usage?

### DIFF
--- a/libcore/src/mpi/LCGrid.h
+++ b/libcore/src/mpi/LCGrid.h
@@ -59,7 +59,7 @@ private:
     double _gridXmin, _gridXmax, _gridYmin, _gridYmax;
     /// for convenience
     /// will be delete in next versions
-    Pedestrian ** _localPedsCopy;
+    std::map<int, Pedestrian *> _localPedsCopy;
     ///total number of pedestrians
     int _nPeds;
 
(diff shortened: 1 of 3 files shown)


Ozaq:
@schroedtert Hmm I tried to check slack but the history is gone already. The setup was from @anna-braun. Maybe she can help.

anna-braun:
> @schroedtert Hmm I tried to check slack but the history is gone already. The setup was from @anna-braun. Maybe she can help.

@Ozaq I have several setups, which do not work. Maybe you talking about one of these? 
[57-17-0.zip](https://github.com/JuPedSim/jpscore/files/4034103/57-17-0.zip)
[57-13.zip](https://github.com/JuPedSim/jpscore/files/4034106/57-13.zip)

libcore/src/mpi/LCGrid.h
```diff
@@ -59,7 +59,7 @@ class LCGrid
     double _gridXmin, _gridXmax, _gridYmin, _gridYmax;
     /// for convenience
     /// will be delete in next versions
-    Pedestrian ** _localPedsCopy;
+    std::map<int, Pedestrian *> _localPedsCopy;
```
zeroset: I think this name is very much misleading, because the local copy is actually not a copy. We should rename this.

schroedtert:
Canceled

PR #649, opened 2020-01-17
Adds output path option to the inifile header

zeroset:
Using `<output path="output_directory" />` the output path can be
set. The default value is `results`. This path can be relative to the
current directory (where the simulation is executed) or absolute.

All result files (trajectories and statistics) will be created in this
directory. Additionally the input files will be copied to this
path. After the simulation this folder contains all files needed to
reproduce the simulation.

Currently the folder will be overwritten if it already exists.


schroedtert:
> From the user's point of view: It works.
> 
> Where is the change/modification in the documentation?

Thanks, documentation is now updated.
@gjaeger could you please have a look if this is enough?

gjaeger:
> > From the user's point of view: It works.
> > Where is the change/modification in the documentation?
> 
> Thanks, documentation is now updated.
> @gjaeger could you please have a look if this is enough?

It's enough.

Issue #303, opened 2019-07-11
Error in sources (when using overlapping sources)


chraibi:
## Summary
When using overlapping sources in larger geometries the program fails with a Segmentation fault in the VelocityModel when accessing the position of the neighbors.

## Inifile + Geometry file to reproduce bug
[Personen_ini.xml](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/uploads/a92b2f4f9efcd5e5af081dd82a0ba96f/Personen_ini.xml)

[Bahnsteige.xml](https://gitlab.version.fz-juelich.de/jupedsim/jpscore/uploads/8a9c8bfc24ae68bf3d264379216a6129/Bahnsteige.xml)

anna-braun:
@chraibi 
The Error changed, it's not a segmentation fault any longer:
![Bildschirmfoto 2019-12-20 um 14 30 24](https://user-images.githubusercontent.com/38530070/72244883-800ccd80-35ef-11ea-808e-23555efb69dd.png)
We do not really know if it is a problem of overlapping sources, as the problem also occurs for single sources. Here are the data which create the current error.
[test-wh-20200113.zip](https://github.com/JuPedSim/jpscore/files/4052810/test-wh-20200113.zip)

chraibi:
Ok, the first question for me is it a source problem at all?

removing the sources gives me the following errors
```
[2020-01-14 21:43:30.153] [error] in AddTarget: calling FinalizeTargetLine
```

anna-braun:
@chraibi I have this error, too (master branch, today 09:30h). I guess this is the floor field router problem from PR #623 (@schroedtert ?).     
When I do the simulation without sources, I have this error as well, but the simulation does not crash as it happens when using one or more sources, so it seems to be a problem of the sources.

schroedtert:
No this is not the FF problem, this is solved with #623. This is the error, that the sources created pedestrians at (0,0) and this point is not inside the room.

schroedtert:
Just a question, do source have the possibility to be distributed in a specific room? I couldn't find this parameter in the documentation.

`<group group_id="901" agent_parameter_id="1" room_id="0" subroom_id="1" number="1" goal_id="901" router_id="1"  />`

anna-braun:
@schroedtert The (sub)room is defined in the group as you show it in the example.    
In the sources you can only give a start position, the (sub)room is taken from the group.
[...]

PR #650, opened 2020-01-17
Delete uneeded source functionalities and clean-up

chraibi:
While generating the positions of the sources, there was
some hacking and trying to adapt their velocity (vector) to their
environment.

This often necessitates a call to the router (so that the desired direction can be
initialized).

However, since the positions are not set yet, the router may choke on these agents
without a meaningful position (defaults to (0,0)).

In this PR the following happens wrt. sources:

- Unused functions are deleted
- Positions of sources are initialized
- *Afterwards*...
- adjust the velocity of the source-agents by making a call to
- `AdjustVelocityUsingWeidmann()`. Within this function, there is a unique call to
  the router.
- Refactoring a for-loop using function `GenerateAgents()`

TODO:

- `AgentsSourcesManager::AdjustVelocityUsingWeidmann()` is not tested in this PR
- Alternatively the function `AgentsSourcesManager::AdjustVelocityByNeighbour()`
  can be used as well.

Related to #303


libcore/src/pedestrian/AgentsSource.cpp
```diff
@@ -73,7 +73,7 @@ AgentsSource::~AgentsSource() {}
 void AgentsSource::GenerateAgentsAndAddToPool(int count, Building * building)
 {
     std::vector<Pedestrian *> peds;
-    GenerateAgents(peds, count, building);
+    GenerateAgents(peds, count, building); // positions are not set yet!
```
schroedtert: Could you add this comment as comment for GenerateAgents in the header file?

libcore/src/pedestrian/AgentsSourcesManager.cpp
```diff
@@ -140,6 +133,10 @@ bool AgentsSourcesManager::ProcessAllSources() const
             } else if(!ComputeBestPositionVoronoiBoost(src.get(), peds, _building, source_peds))
                 LOG_WARNING("There was no place for some pedestrians");
 
+            // Having set the positions, now we can set the velocity
+            for(auto ped : peds)
+                AdjustVelocityUsingWeidmann(ped);
+
```
schroedtert: Please add braces:
```
for(auto ped : peds){
    AdjustVelocityUsingWeidmann(ped); 
}
```

libcore/src/pedestrian/AgentsSource.cpp
```diff
@@ -220,27 +220,20 @@ const std::shared_ptr<StartDistribution> AgentsSource::GetStartDistribution() co
...
+    //          the positions will be calculated and initializes.
+    // TODO: We should reverse the order of points 1 and 2.
     int pid;
     pid = (this->GetAgentId() >= 0) ?
               this->GetAgentId() :
               Pedestrian::GetAgentsCreated() + building->GetAllPedestrians().size();
     for(int i = 0; i < count; i++) {
         if(GetStartDistribution()) {
             auto ped = GetStartDistribution()->GenerateAgent(building, &pid, emptyPositions);
-            if(ped->FindRoute() == -1) {
```
schroedtert: `FindRoute` is now only called after the position is set?

libcore/src/pedestrian/AgentsSourcesManager.cpp
```diff
@@ -53,12 +53,7 @@ void AgentsSourcesManager::Run()
     LOG_INFO("Starting agent manager thread");
     //Generate all agents required for the complete simulation
     //It might be more efficient to generate at each frequency step
-    //TODO  this loop is exactly GenerateAgents( --> REFACTOR)
-    for(const auto & src : _sources) {
-        LOG_INFO("Generate src: {}", src->GetId());
-        src->GenerateAgentsAndAddToPool(src->GetMaxAgents(), _building);
-    }
-
+    GenerateAgents();
```
schroedtert: Are all pedestrian computed beforehand or only the ones added in this step? Is this comment still valid?

libcore/src/pedestrian/AgentsSource.cpp
chraibi: ok

libcore/src/pedestrian/AgentsSourcesManager.cpp
chraibi: ok

libcore/src/pedestrian/AgentsSource.cpp
chraibi: Yes, in `AdjustVelocityUsingWeidmann(ped)` called in `AgentsSourcesManager.cpp`

libcore/src/pedestrian/AgentsSourcesManager.cpp
chraibi: This is just refactoring. Functionality was not altered. 
Here, for every source a `MaxAgents` are generated and added to Pool, waiting for `processAllSources` to be added to `building`.

PR #654, opened 2020-01-25
Clean up EventManager

schroedtert:
With this PR the functionality of the EventManager reduced. It now only can process the Event (e.g., open, close, temp_close, reset transitions). The reading of the event and schedule file was moved to a new EventFileParser, where additional checks during parsing were added.
Also the knowledge spreading is removed completely from the EventManager as it was broken at the moment. It may added back, with a specific KnowledgeManager later. This leads to the removal of the following input options in the event file: `update_frequency`, `update_radius`, and `agents_color_by_knowledge`, `caption`.
The field `type` from the events is also removed as it is always set to `door`.

### DIFF
--- a/demos/scenario_5_events/events_ini.xml
+++ b/demos/scenario_5_events/events_ini.xml
@@ -59,9 +59,9 @@ xsi:noNamespaceSchemaLocation="http://xsd.jupedsim.org/0.6/jps_ini_core.xsd">
   </routing>
 
   <!--persons information and distribution -->
-  <agents operational_model_id="2">
+  <agents operational_model_id="1">
     <agents_distribution>
-      <group group_id="0" agent_parameter_id="1" room_id="0" subroom_id="0" number="1" router_id="1" />
+      <group group_id="0" agent_parameter_id="1" room_id="0" subroom_id="0" number="1" router_id="2"/>
       <!-- <group group_id="1" agent_parameter_id="1" room_id="0" subroom_id="1" number="20" router_id="1" /> -->
       <!-- <group group_id="2" agent_parameter_id="1" room_id="0" subroom_id="2" number="4" router_id="1" /> -->
       <!-- <group group_id="3" agent_parameter_id="1" room_id="0" subroom_id="3" number="20" router_id="1" /> -->
--- a/demos/scenario_5_events/events_ini.xml
+++ b/demos/scenario_5_events/events_ini.xml
@@ -61,7 +61,7 @@ xsi:noNamespaceSchemaLocation="http://xsd.jupedsim.org/0.6/jps_ini_core.xsd">
   <!--persons information and distribution -->
   <agents operational_model_id="1">
     <agents_distribution>
-      <group group_id="0" agent_parameter_id="1" room_id="0" subroom_id="0" number="1" router_id="2"/>
+      <group group_id="0" agent_parameter_id="1" room_id="0" subroom_id="0" number="1" router_id="1"/>
       <!-- <group group_id="1" agent_parameter_id="1" room_id="0" subroom_id="1" number="20" router_id="1" /> -->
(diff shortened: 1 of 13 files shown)


demos/scenario_5_events/events_ini.xml
```diff
@@ -59,9 +59,9 @@ xsi:noNamespaceSchemaLocation="http://xsd.jupedsim.org/0.6/jps_ini_core.xsd">
   </routing>
 
   <!--persons information and distribution -->
-  <agents operational_model_id="2">
```
zeroset: Why did you change router and model here?

libcore/src/IO/EventFileParser.h
```diff
@@ -0,0 +1,51 @@
...
+ **/
+#pragma once
+
+#include "events/Event.h"
+#include "general/Filesystem.h"
+
+#include <map>
+#include <vector>
+
+class EventFileParser
```
zeroset: Can we use a namespace here? All methods are static anyways.

libcore/src/IO/IniFileParser.cpp
```diff
@@ -165,14 +165,6 @@ void IniFileParser::Parse(const fs::path & iniFile)
 
 bool IniFileParser::ParseHeader(TiXmlNode * xHeader)
 {
-    LOG_INFO("JuPedSim - JPScore");
```
zeroset: Why did you delete this? Shouldn't we just move it somewhere else?

libcore/src/Simulation.cpp
```diff
@@ -187,15 +188,19 @@ bool Simulation::InitArgs()
         return false;
     }
 
-    _em = new EventManager(_config, _building.get(), _config->GetSeed());
-    if(!_em->ReadEventsXml()) {
-        LOG_WARNING("Could not initialize events handling");
+    _em = new EventManager(_building.get());
```
zeroset: Does it need to be a pointer?

demos/scenario_5_events/events_ini.xml
schroedtert: Operational model 2 is no longer supported. And router is changed back to `global_shortest`.

libcore/src/IO/EventFileParser.h
schroedtert: Changed to namespace.

libcore/src/IO/IniFileParser.cpp
schroedtert: Is already printed in jpscore/src/main.cpp:
```    LOG_INFO("Starting JuPedSim - JPScore");
    LOG_INFO("Version {}", JPSCORE_VERSION);
    LOG_INFO("Commit id {}", GIT_COMMIT_HASH);
    LOG_INFO("Commit date {}", GIT_COMMIT_DATE);
    LOG_INFO("Build from branch {}", GIT_BRANCH);
    LOG_INFO("Build with {}({})", compiler_id, compiler_version);

File: libcore/src/Simulation.cpp (hunk 4 above)
Comment (schroedtert) [reply libcore/src/Simulation.cpp]: Changed it to std::unique_ptr<EventManager>

File: libcore/src/IO/EventFileParser.h (hunk 2 above)
Comment (schroedtert) [reply libcore/src/IO/EventFileParser.h]: Changed to namespace.

File: libcore/src/IO/EventFileParser.cpp [hunk 5]
```diff
@@ -0,0 +1,317 @@
...
+        double time;
+        if(const char * attribute = e->Attribute("time"); attribute) {
+            if(double value = xmltof(attribute, std::numeric_limits<double>::min()); value >= 0.) {
+                time = value;
+            } else {
+                LOG_ERROR("event {:d}: time not set properly", id);
+                continue;
+            }
+        } else {
+            LOG_ERROR("event {:d}: time required", id);
```
zeroset: `{:d}` is the default for formatting integers. I am not sure if we should keep the format specifier or use empty braces

PR #656, opened 2020-01-27
Add dockerfile for DockerHub

chraibi:


container/dockerhub/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+                       make \
+                       libboost-dev \
+                       libcgal-dev \
+                       clang-format-8 \
+                       python3 \
+                       python3-pip \
+                       python \
+                       python-pip
+
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6084F3CF814B57C1CF12EFD515CF4D18AF4F7421 && \
```
zeroset: Do you need clang-format-9?

container/dockerhub/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN apt-get update && apt-get install -y software-properties-common
+
+RUN apt-get update && apt-get install -y wget \
+                       git \
+                       g++-8 \
+                       cmake \
+                       make \
+                       libboost-dev \
+                       libcgal-dev \
+                       clang-format-8 \
```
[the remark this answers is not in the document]
chraibi: no

container/dockerhub/Dockerfile
chraibi: no

container/dockerhub/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y software-properties-common
```
zeroset: I think this was required for adding the other repository for clang-format 9 and should be removed as well.

container/dockerhub/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    && git clone --depth=1 https://github.com/JuPedSim/jpscore.git \
+    && cp jpscore/scripts/setup-deps.sh /opt \
+    && cd /opt && CXX=/usr/bin/g++-8 CC=/usr/bin/gcc-8 ./setup-deps.sh
+
+# build jpscore and jpsreport
+RUN cd /home/jupedsim/workspace/jpscore \
+    && mkdir -p build \
+    && cd build \
+    && cmake  ..\
+    && make
```
zeroset: Maybe you should use multithreading here `-j$(nproc)` ?

container/dockerhub/Dockerfile
chraibi: ok

container/dockerhub/Dockerfile
chraibi: Yes, thanks!

Ozaq:
I don't think this is the right solution, if i understand this correctly this is supposed to be a usage container not a build container.

As a user I would be quite surprised (annoyed) if I had to download such a fat container. I think the image should not build the software but instead just copy it in from TC. Something like:

```
FROM ubuntu:18:04
COPY <jps executable>
...
```

Issue #657, opened 2020-01-28
Interface for XML parsing


schroedtert:
Create a better interface for parsing specific values from a xml file.

AC:
- templated interface which interacts with TinyXml nodes and throws expressive exceptions
- library for parsing strings to desired data type:
    - check if input is datatype
    - check if in datatype range
- data types needed:
    - integer (signed/unsigned)
    - double (signed/unsigned)
    - bool
    - fs::path
- unit tests
- optional: add conditions to parsing (e.g., allowed range for numerical values)

schroedtert:
Instead of implementing a new parser it might be easier to use the correct functions from tinyxml:
https://leethomason.github.io/tinyxml2/classtinyxml2_1_1_x_m_l_element.html#afea43a1d4aa33e3703ddee5fc9adc26c

PR #658, opened 2020-01-28
XML parsing interface

schroedtert:
As described in #657, a more comfortable interface for reading specific values from XML documents is developed.

Comments/remarks for improvement are greatly appreciated!


schroedtert:
To simplify things, we also could use tinyxml correctly:
https://leethomason.github.io/tinyxml2/classtinyxml2_1_1_x_m_l_element.html#afea43a1d4aa33e3703ddee5fc9adc26c

Issue #461, opened 2019-08-08
Changing the binary output dir to <build>/bin/ broke all tests based on "JPSRunTestDriver"


Ozaq:
Ok this is my fault :/ JPSRunTestDriver does some path adjustment and assumes `jpscore` always to be in `<source>/bin/jpscore`. I foolishly thought the test might not run because of different reasons.

I am working on a patch, hope to have it available soon.

gjaeger:
@Ozaq If you have an idea how we can simplify `JPSRunTestDriver`, this would be great. 

A good solution could also be helpful here ([JPSreport, issue 130](https://github.com/JuPedSim/jpsreport/issues/130)).

Ozaq:
@gjaeger W.r.t. the path to `jpscore`, I would like to make this a test argument, which then can be set from CMake `add_test()`.

One thing that needs discussion are the output files. I would like to generate them into the current working directory or in a specified path, e.g. `jpscore --output-path=...`, but given my current understanding this would make it necessary to generate a relative path into the generated `trajectory.xml` files. Thus making it impossible to relocate them.

I would prefer if results where relocatable and possible to visualize w.o. the input data but this is not required to address this issue.

gjaeger:
> One thing that needs discussion are the output files. I would like to generate them into the current working directory or in a specified path, e.g. jpscore --output-path=..., but given my current understanding this would make it necessary to generate a relative path into the generated trajectory.xml files. Thus making it impossible to relocate them.

That's a good idea.

**JPSvis** already offers the possibility to load a trajectory file without geometry. The geometry can then be selected separately. We can extend the routine so that the geometry can be loaded manually if the path is incorrect.

PR #667, opened 2020-02-02
Automatically set jpscore executable for systemtest

schroedtert:
Instead of using `build/bin/jpscore` a jpscore executable can be passed as argument to JPSRunTest. For the systemtests this executable is automatically set to the executable build with the build command.
See #461.

### DIFF
--- a/systemtest/CMakeLists.txt
+++ b/systemtest/CMakeLists.txt
@@ -1,3 +1,6 @@
+# get used jpscore executable
+set(jpscore_exe $<TARGET_FILE:jpscore>)
+
 add_subdirectory(test_geometry)
 add_subdirectory(juelich_tests)
 add_subdirectory(rimea_tests)
--- a/systemtest/JPSRunTest.py
+++ b/systemtest/JPSRunTest.py
@@ -27,17 +27,18 @@ def getScriptPath():
 
 class JPSRunTestDriver(object):
 
-    def __init__(self, testnumber, argv0, testdir, utestdir="..", jpsreportdir=""):
+    def __init__(self, testnumber, argv0, testdir, utestdir="..", jpsreportdir="", jpscore=""):
         self.SUCCESS = 0
         self.FAILURE = 1
         # check if testnumber is digit
         assert isinstance(testnumber, float) or isinstance(testnumber, int), "argument <testnumber> is not digit"
         # only allow path and strings as path directory name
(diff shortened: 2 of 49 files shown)


systemtest/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+# get used jpscore executable
+get_property(jpscore_name TARGET jpscore PROPERTY NAME)
```
zeroset: Maybe also check if `OUTPUT_NAME` is empty, that could be useful if we set the name by hand.

systemtest/JPSRunTest.py
```diff
@@ -67,8 +69,9 @@ def __init__(self, testnumber, argv0, testdir, utestdir="..", jpsreportdir=""):
     def run_test(self, testfunction, fd=0, *args): #fd==1: make fundamental diagram
         assert hasattr(testfunction, '__call__'), "run_test: testfunction has no __call__ function"
         self.__configure()
-        jpscore = os.path.join(self.trunk, "build", "bin", "jpscore")
-        jpscore_exe = self.__find_executable(jpscore)
+        jpscore_exe = self.__find_executable(self.jpscore)
```
zeroset: Do we really use find_executable? shouldn't we remove it and make an assert that the file given in jpscore exits?

chraibi: No, we don't really need it.
This testing machinery is supposed to tun under Linux. So this "windows hack" is not necessary.

zeroset: Depends, the test should also run on windows as we want to test our binaries on windows too.

chraibi: Let's keep it this way. This Driver needs some refactoring, now that jpsreport needs to be tested as well.
We probably should open another issue/PR for this.

systemtest/CMakeLists.txt
schroedtert: Changed to generator expression

systemtest/JPSRunTest.py
schroedtert: From my understanding is `find_executable` not needed, as the executable is given to the script. The function searches  for `jpscore.exe` in the old binary output directory which may be empty.

chraibi: It would be better to retrieve these variables from CMAKE. 
This driver needs refactoring anyway.
Do you want to do it in this PR?

PR #671, opened 2020-02-07
remove subprocess call in test driver

### LINKED ISSUE #568: Python script calls python as subprocess
In `JPSRunTestDriver` the makeinifile.py script is calle using 
```subprocess.call(["python", "makeini.py", "-f", "%s" % self.FILE])``` 
Depending on the system this is python2 or python3. We should stick to python3 and avoid calling command line programs like this. The user calls the python script using a specific python version and this should be used! 
That is the reason why we have to install numpy in python2 and python3 at the moment.

### PR DESCRIPTION
**Pull request by chraibi**
- Remove subprocess call to makeini.py

fixes #568

### CODE REVIEW

**Comment by schroedtert** [comment]
See #568

File: systemtest/juelich_tests/test_11/runtest_juelich_11.py [hunk 1]
```diff
@@ -1,78 +1,45 @@
...
+def get_evak_time(inifile, trajfile):
+    fps, N, traj = parse_file(trajfile)
+    etime =  (max( traj[:, 1] ) - min(traj[:, 1])) / float(fps)
+    return etime
 
+def call_test(num, suffix):
+    logging.info("copy master_ini_%s.xml"%suffix)
+    logging.info("copy geometry/geometry_test11_%s.xml"%suffix,)
+    copyfile("%s/master_ini_%s.xml"%(PATH, suffix), "%s/master_ini.xml"%PATH)
+    copyfile("%s/geometry/geometry_test11_%s.xml"%(PATH, suffix), "%s/geometry.xml"%PATH)
```
schroedtert: Why are these files copied? Do we need them somewhere else? Isn't this done by make_ini?

systemtest/juelich_tests/test_14/master_ini.xml
```diff
@@ -2,7 +2,7 @@
 
 <JuPedSim project="JPS-Project" version="0.8" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="ini.xsd">
 
-  <seed>linspace(1, 10000, 100)</seed>
+  <seed>np.linspace(1, 10000, 100)</seed>
```
schroedtert: Do we need this np here?

systemtest/juelich_tests/test_3/master_ini.xml
```diff
@@ -3,7 +3,7 @@
 <JuPedSim project="Test3" version="0.8" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="ini.xsd">
 
   <!-- seed used for initialising random generator -->
-  <seed>linspace(1, 100000, 20)</seed>
+  <seed>np.linspace(1, 100000, 20)</seed>
```
schroedtert: Is the np needed?

systemtest/juelich_tests/test_11/runtest_juelich_11.py
chraibi: these are copied because this test is composed of two subtests that cover two cases:
- Subrooms
- Rooms

Idially makeini, would glob over geometries, but I don't know how to do neither I would like to do it, knowing that makeini has no future.

systemtest/juelich_tests/test_14/master_ini.xml
chraibi: Yes, otherwise the test fails.

systemtest/juelich_tests/test_3/master_ini.xml
chraibi: Yes, otherwise the test fails.

chraibi:
This PR depends on:

-  #704
- #703

PR #672, opened 2020-02-09
Add a systemtest based on pytest

Ozaq:
A new set of test helper functions / class have been added as python
module under 'systemtest/jps'. Please see the individual files for
details about their functionality.

The newly added systemtest can be found in
'systemtest/test_comandline_args.py', each 'test_*' function represents
a test case.

For now the tests are not hooked up to cmake and can be executed with:
'JPSCORE=<path to your jpscore binary> pytest -s -vv'

Note you will need pytest installed in your python environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -192,3 +192,7 @@ titlepage.tex
 
 __pycache__
 ipynb_checkpoints/
+
+# Ignore vscode workspace files
+*.code-workspace
+.vscode
--- a/jpscore/src/main.cpp
+++ b/jpscore/src/main.cpp
@@ -67,7 +67,6 @@ int main(int argc, char ** argv)
         iniFileParser.Parse(a.IniFilePath());
     } catch(const std::exception & e) {
         LOG_ERROR("Exception in IniFileParser::Parse thrown, what: {}", e.what());
-
         return EXIT_FAILURE;
     }
 
--- a/systemtest/jps/checker.py
+++ b/systemtest/jps/checker.py
@@ -1,0 +1,14 @@
+"""Functions to check jpscore log files for specific output."""
+
+
+def output_contains_line_exact(text, line):
+    """
+    Search for a specific line in the output.
+
+    @param text to search in. This is the output of jpscore usually.
+    @param line to search for.
+    """
+    for line in text.splitlines():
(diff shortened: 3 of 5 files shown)


Ozaq:
@zeroset @chraibi @schroedtert and @gjaeger I got a bit fed up with the system tests and created an example of how i think this should look like. Please have a look and tell me if this helps. Cheers :)

gjaeger:
@Ozaq What do you mean with "_system tests_"? Do you mean test cases for checking the coding/programming?

Various terms are used to check pedestrians models. In the field of fire simulation the terms _validation_ and _verification_ are used for different aspects. 

flowchart_V-V_EN

How do you see it from the perspective of a computer scientist?

Ozaq:
@gjaeger ah :) welcome to the murky world of software testing!

 I used the imo most appropriate ISTQB term, see: https://glossary.istqb.org/en/search/system%20testing

So yes this encompasses verification but also includes other aspects of your software as well, e.g. non-functional criteria such as response time, memory usage but also "are my comand line options working"

Ozaq:
@chraibi thx!
About your questions:
Q: How do you intend to include jpsreport in the testing machinery?
A: I think I would add another class to `driver.py` named JpsReport that deals with report. Some common functionality could be extracted. Should be pretty straight forward to do so.

Q: What about calling makeini.py to produce several inifiles?
A: That can be done, for the test cases I wanted to cover this was just no yet necessary. 
Although I would probably use `jina2` to generate the files, instead of a home-cooked solution: See https://jinja.palletsprojects.com/en/2.11.x/

chraibi:
`jina2` looks nice and offers at least the same functionalities as our fabulous makeini-script. ^^

.clang-tidy
```diff
@@ -1,5 +1,12 @@
 ---
-Checks:          'clang-diagnostic-*,clang-analyzer-*,*, -fuchsia*, -readability-isolate-declaration, -cppcoreguidelines-avoid-magic-numbers, -readability-magic-numbers,-modernize-use-trailing-return-type'
+Checks:>
```
zeroset: Love this :>

zeroset:

PR #700, opened 2020-02-19
Add code coverage to unittest

schroedtert:
Added new cmake option CODE_COVERAGE to add code coverage analysis to unittests. Only supported with Unix/MacOS, needs gcc or clang compiler. 
Following libraries are needed:
-  gcc: `lcov`
- clang: `llvm-cov `and `llvm-profdata`


cmake_modules/code-coverage.cmake
```diff
@@ -0,0 +1,559 @@
+#
```
zeroset: Could you add a comment with an url where this is from? And maybe which version/git hash?

schroedtert: added

PR #704, opened 2020-02-21
Remove juelich tests 10 15

chraibi:
Remove juelich tests 10 and 15.


zeroset:
Why did you remove the Tests?

chraibi:
> Why did you remove the Tests?

- Test 15: Because this test is not really testing the speed of the software. I mean,  `making sure that the code is not getting slower` is probably not easy to measure. Besides, the test relies on ta log file, that does not exist.

- Test 10: According to the documentation tests Verlet Lists. I don't think there is something left in the code of Verlet lists. I think, in the future when we address the LC-Grid, more specific tests need to be written.

PR #711, opened 2020-03-02
Add results folder in demos to gitignore and exclude them from install

schroedtert:
When executing any of the demos and then rerunning cmake always led to:
```
CMake Error at jpscore/CMakeLists.txt:70 (install):
  install FILES given directory
  "/home/.../jpscore/demos/01_corridor/results" to install.
```
Now the each file needs to be mentioned explicitly to be copied to the install path.
Additionally all the files in the default output results for the demos was added to gitignore, to avoid unintentional uploads.

### DIFF
--- a/jpscore/CMakeLists.txt
+++ b/jpscore/CMakeLists.txt
@@ -66,7 +66,34 @@ install(TARGETS jpscore
         COMPONENT applications)
 
 set(CT_DATA_FILE_DIR "demos")
-file(GLOB CT_FILES "${CMAKE_SOURCE_DIR}/${CT_DATA_FILE_DIR}/*/*")
+set(CT_FILES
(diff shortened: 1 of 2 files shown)


jpscore/CMakeLists.txt
```diff
@@ -66,7 +66,34 @@ install(TARGETS jpscore
         COMPONENT applications)
 
 set(CT_DATA_FILE_DIR "demos")
-file(GLOB CT_FILES "${CMAKE_SOURCE_DIR}/${CT_DATA_FILE_DIR}/*/*")
+set(CT_FILES
```
chraibi: Hey, why do you write these file explicitly?

schroedtert: To make sure that only the files are copied to the install path that are defined. Otherwise all files within the demo directory would be copied even if not intended by us, e.g., the user has added some files to the directory.

PR #714, opened 2020-03-05
Update CDFdistance.py

Shuchao-git:
According to the definition of validity factor, there is a mistake in the calculation in line 102. I corrected it in the new file.


zeroset:
I did not forget about it, have to configure something on TeamCity... sorry for delay.

Issue #436, opened 2019-07-21
Remove xsd Files. 


chraibi:
**Is your feature request related to a problem? Please describe.**
We have some `xsd` files, however, there are not up to date. 

Unless we can find an easy way to validate our XML files using these XSD files, I don't see the point in keeping them around. 

Any ideas?

Ozaq:
XSD files are the specs for your XML files and as such are invaluable information when trying to parse or generate those XML files. I would suggest to update them and move from tinyxml to xerceces (https://xerces.apache.org/xerces-c/) Xerxes has a build in validation ox XML files versus their XSD. If I remember correctly the generated error messages are solid and allow the creator of an XML file to quickly pinpoint the error, e.g. missed an required attribute, supplied a double where an int was required etc...

This would greatly simplify the parsing code since a lot less manual error checking would be necessary.

Ozaq:
Looking trough the issues its things like #382  #161 that probably would be a lot less hassle if xml validation with the accompanying xsd was used.

chraibi:
I'm with you that validating of xml files with xsd will save us a lot of trouble. 
I'll have a look at xerces. It looks like it can do what we want..

Ozaq:
I worked some with Xerxes in the past and it was pretty good, but i think migration would be a significant effort due to the difference in API. Still worth it though imo.

zeroset:
We will not have an XSD validating XML parser. We should delete all xsd files and remove them from the xml files.

chraibi:
no objetcions

PR #716, opened 2020-03-07
Remove xsd files

schroedtert:
We do not use the xsd for xml validation and they are outdated, so they are removed.
- Remove xsd files
- Remove xsd calls in demo files

Solves #436


zeroset:
Please also remove from systemtests ;)

schroedtert:
> Please also remove from systemtests ;)

Hopefully now everything is removed.

PR #717, opened 2020-03-09
Remove unneeded inifile option solver

schroedtert:
As we only support Euler solver at the moment the option is always set to euler and hence not needed.
Additionally the parsing of solver always raised an error message although no error occurs.
- Removes solver option from parsing
- Removes solver from configuration
- Removes option from documentation
- Removes option from demo files


schroedtert:
> Please also remove the solver from the systemtests (e.g. correct geometry tests)

Hopefully now everything is gone

PR #721, opened 2020-03-12
Refactor UpdateRoutesAndLocations

schroedtert:
Instead of one large function the function is split into several smaller functions by their functionality and moved to a helper namespace SimulationHelper:
- UpdateLocations: Give a list of pedestrians who changed the room, or moved in an unusual matter. Uses UpdateRoom
- UpdateRoom: Checks if the pedestrian has changed its room/subroom, if so update the room information. If not in a neighboring room/subroom report error.
- FindOutsidePedestrians: Find all pedestrians that have left the geometry
- UpdateFlowAtDoors: Find the passed door of a pedestrian and increases its usuage
- UpdateFlowRegulation: Trigger flow regulation of transition if needed
- FindPassedDoor: Find the door, which was passed by a pedestrian in the last time step

Following parts have been moved to seperate Function in Simulation:
- RemovePedestrians: Remove all pedestrians which are marked for removing, e.g. left the building in a normal way or changed their position in an unusual way.
- UpdateRoutes: Updates the routes based on the changes (geometrical and positional) of the current compute step

Following was moved to the main loop
- GoalManager::Process

Following work still needs to be done:

- [x] unittests
- [x] ~~integration tests~~
- [x] refactoring, in particular check if these functions need to be member functions of Simulations or could be free functions
- [x] clean up documentation
- [ ] discussion about program flow


libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+ **/
+#include "SimulationHelper.h"
+
+#include "general/Logger.h"
+#include "geometry/Room.h"
+#include "geometry/SubRoom.h"
+
+#include <algorithm>
+
+std::optional<bool> SimulationHelper::UpdateRoom(const Building & building, Pedestrian & ped)
```
zeroset, line 28: Do you really need optional of bool? I think this might be error prone since optional can be implicit converted to bool as well. see https://www.boost.org/doc/libs/1_57_0/libs/optional/doc/html/boost_optional/tutorial/a_note_about_optional_bool_.html

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+    auto oldSubRoom = building.GetRoom(ped.GetRoomID())->GetSubRoom(ped.GetSubRoomID());
+    if(oldSubRoom->IsInSubRoom(&ped)) {
+        return false;
+    }
+
+    // relocation needed, look if pedestrian is in one of the neighboring rooms/subrooms
+    // get all connected directly connected rooms/subrooms
+    // TODO check for closed doors?
+    std::map<int, std::shared_ptr<SubRoom>> subroomsConnected;
+    for(const auto & [roomID, room] : building.GetAllRooms()) {
```
zeroset, line 41: The `SubRoom` has a variable `_neighbors` which seems to be containing all neighboring sub-rooms. Maybe this one should be used, an additional  method might be required on `SubRoom`.

libcore/src/SimulationHelper.cpp
zeroset, line 28: Maybe rename the function. At the moment it sounds like the room is updated, but the pedestrians room is updated.

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+    const std::vector<Pedestrian *> & peds)
+{
+    std::vector<Pedestrian *> pedsAtFinalGoal;
+    const auto & goals = building.GetAllGoals();
+
+    std::copy_if(
+        std::begin(peds),
+        std::end(peds),
+        std::inserter(pedsAtFinalGoal, std::end(pedsAtFinalGoal)),
+        [goals](auto const ped) -> bool {
```
zeroset, line 82:
```suggestion
        [goals](const Pedestrian * ped) -> bool {
```
I thinks auto is a bit missleading when you get a pointer. And we want the Pedestrian const and not the pointer.

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+            return ped->GetFinalDestination() != FINAL_DEST_OUT &&
+                   goals.at(ped->GetFinalDestination())->Contains(ped->GetPos()) &&
+                   goals.at(ped->GetFinalDestination())->GetIsFinalGoal();
+        });
+    return pedsAtFinalGoal;
+}
+
+std::tuple<std::vector<Pedestrian *>, std::vector<Pedestrian *>>
+SimulationHelper::UpdateLocations(const Building & building, const std::vector<Pedestrian *> & peds)
```
zeroset, line 91: Which Locations are updated? This is very hidden in the UpdateRoom Method. We should rename the function.

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+            }
+        } else {
+            pedsNotRelocated.push_back(ped);
+        }
+    }
+
+    return {pedsChangedRoom, pedsNotRelocated};
+}
+
+std::vector<Pedestrian *> SimulationHelper::FindOutsidePedestrians(
```
zeroset, line 110:
```suggestion
std::vector<Pedestrian *> SimulationHelper::FindPedestriansOutside(
```

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+    const Building & building,
+    std::vector<Pedestrian *> & peds)
+{
+    std::vector<Pedestrian *> pedsOutside;
+    std::vector<Pedestrian *> newPeds;
+
+    std::partition_copy(
+        std::make_move_iterator(std::begin(peds)),
+        std::make_move_iterator(std::end(peds)),
```
zeroset, line 119: Why do you use move here?

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+            return transPassed.value()->IsExit() && transPassed.value()->IsOpen() &&
+                   transPassed.value()->DistTo(ped->GetPos()) < maxDistance;
+        });
+    peds = std::move(newPeds);
+    return pedsOutside;
+}
+
+void SimulationHelper::UpdateFlowAtDoors(
+    Building & building,
+    const std::vector<Pedestrian *> & peds)
```
zeroset, line 139: Maybe this should be called `peds_changed_room` to be clear that this function is called for some and not for all pedestrians...

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+        if(!closestTransition.has_value()) {
+            LOG_WARNING("Ped {} did not cross any transition", ped->GetID());
+            return;
+        }
+
+        //TODO check if exitindex is set and transition exists?
+        if(closestTransition.value()->GetUniqueID() !=
+           building.GetTransitionByUID(ped->GetExitIndex())->GetUniqueID()) {
+            LOG_WARNING(
+                "Ped {}: used an indented door {}, but wanted to go to {}.",
```
zeroset, line 153: This sentence is either wrong or too confusing. didn't the pedestrian use an unintended door?

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+std::optional<Transition *>
+SimulationHelper::FindPassedDoor(const Building & building, const Pedestrian & ped)
+{
+    std::vector<Transition *> transitions;
+
+    // TODO check if room exists?
+    auto room = (ped.GetRoomID() != -1) ? (building.GetAllRooms().at(ped.GetRoomID())) :
+                                          (building.GetAllRooms().at(ped.GetOldRoomID()));
+
+    for(auto const & [_, subroom] : room->GetAllSubRooms()) {
```
zeroset, line 174: Are you sure, `_` does not produce a warning on the supported compilers?

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+            std::end(transitions),
+            std::begin(subroom->GetAllTransitions()),
+            std::end(subroom->GetAllTransitions()));
+    }
+
+    Line step{ped.GetLastPosition(), ped.GetPos()};
+    // TODO check for closed doors and distance?
+    auto passedTrans = std::find_if(
+        std::begin(transitions), std::end(transitions), [&step](auto const & trans) -> bool {
+            return trans->IntersectionWith(step) == 1;
```
zeroset, line 185: What is in the case that the lines somehow overlap?

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+bool SimulationHelper::UpdateFlowRegulation(Building & building)
+{
+    bool stateChanged = false;
+
+    for(auto [transID, trans] : building.GetAllTransitions()) {
+        DoorState state = trans->GetState();
+
+        bool regulateFlow = trans->GetOutflowRate() < (std::numeric_limits<double>::max)() ||
```
zeroset, line 202: are those brackets `(std::numeric_limits<double>::max)()` on purpose?

libcore/src/SimulationHelper.cpp
schroedtert: Changed return type to an enum

libcore/src/SimulationHelper.cpp
schroedtert: Totally overlooked this variable, thanks! Used it now.

libcore/src/SimulationHelper.cpp
schroedtert: Changed the name

libcore/src/SimulationHelper.cpp
schroedtert: Updated the function name

libcore/src/SimulationHelper.cpp
schroedtert: Misunderstanding from my side. Changed it.

libcore/src/SimulationHelper.cpp
schroedtert: Changed the name

libcore/src/SimulationHelper.cpp
schroedtert: Should be unintended, changed it

libcore/src/SimulationHelper.cpp
schroedtert: Changed the name to `subroomID`

libcore/src/SimulationHelper.cpp
schroedtert: No, copy paste from former implementation, removed them

libcore/src/SimulationHelper.cpp
schroedtert: The function now also deals with the cases:
1. Current step ends on line: No passed door
2. Current step is on line: No passed door
3. Current step starts on line: Door passed
Added the corresponding tests

libcore/src/Simulation.cpp
```diff
@@ -214,93 +215,55 @@ double Simulation::RunStandardSimulation(double maxSimTime)
...
+
+    // TODO not needed at the moment, as we do not have inside final goals yet
+    auto pedsAtFinalGoal = SimulationHelper::FindPedsReachedFinalGoal(*_building, peds);
+    _pedsToRemove.insert(pedsAtFinalGoal.begin(), pedsAtFinalGoal.end());
+
+    auto pedsOutside = SimulationHelper::FindPedestriansOutside(*_building, pedsNotRelocated);
+    _pedsToRemove.insert(pedsNotRelocated.begin(), pedsNotRelocated.end());
+    _pedsToRemove.insert(pedsOutside.begin(), pedsOutside.end());
+
+    //TODO see if this can be merged
```
zeroset, line 234: Can remove some of the TODO comments before merging?

libcore/src/Simulation.cpp
```diff
@@ -214,93 +215,55 @@ double Simulation::RunStandardSimulation(double maxSimTime)
...
+    SimulationHelper::UpdateFlowAtDoors(*_building, pedsChangedRoom);
+    SimulationHelper::UpdateFlowAtDoors(*_building, pedsOutside);
+
+    //TODO free function?
+    RemovePedestrians();
```
zeroset, line 239: Can we improve the name even further? Which pedestrians are removed?

zeroset: And can you make it a free function with an argument which pedestrians should be removed and where?

libcore/src/Simulation.cpp
```diff
@@ -383,6 +332,7 @@ void Simulation::PrintStatistics(double simTime)
         }
     }
 
+    //TODO discuss deletion
```
zeroset, line 335: Should this todo be merged into master or was it a comment for review?

libcore/src/SimulationHelper.cpp
```diff
@@ -0,0 +1,207 @@
...
+        ped.SetSubRoomUID((*currentSubRoom)->GetUID());
+        return PedRelocation::SUCCESSFUL;
+    } else {
+        ped.UpdateRoom(-1, -1);
+        ped.SetSubRoomUID(-1);
+        return PedRelocation::FAILED;
+    }
+}
+
+std::vector<Pedestrian *> SimulationHelper::FindPedsReachedFinalGoal(
```
zeroset, line 59: We should be more consistent on Ped vs Pedestrian and singular and plural.

schroedtert: - Pedestrian: In function name when only used on single pedestrian
- Pedestrians: In function name when used on multiple/all pedestrian
- ped: In parameter name if function works on single pedestrian
- peds: In parameter name if function works on multiple pedestrians

libcore/src/Simulation.cpp
schroedtert: Todo deleted

libcore/src/Simulation.cpp
schroedtert: Split into free functions in SimulationHelper:
 - RemovePedestrians, for all regular pedestrians to remove, e.g., reached outside or final goal
 - RemoveFaultyPedestrians, for all pedestrians where issues occur, currently pedestrians who moved in an unexpected manner (not to a neighboring room). Could also be used to track pedestrian with issues from the models.

libcore/src/Simulation.cpp
schroedtert: Removed

PR #729, opened 2020-04-03
Fix application suggestions for LGPL 

zeroset:
I wanted to add the license files and harmonize the license notice and copyright notices.


zeroset:

PR #736, opened 2020-04-08
Remove Trajectory XML output format

zeroset:
The XML format for Trajectories should be removed. The Implementation is removed and the systemtest are changed correspondingly. The python parsing of trajectory files is changed to txt format.


libcore/src/IO/IniFileParser.cpp
```diff
@@ -235,12 +235,10 @@ bool IniFileParser::ParseHeader(TiXmlNode * xHeader)
...
         std::string format = xHeader->FirstChildElement("trajectories")->Attribute("format") ?
                                  xHeader->FirstChildElement("trajectories")->Attribute("format") :
-                                 "xml-plain";
+                                 "plain";
         std::transform(format.begin(), format.end(), format.begin(), ::tolower);
 
-        if(format == "xml-plain") {
-            _config->SetFileFormat(FileFormat::XML);
-        } else if(format == "plain") {
+        if(format == "plain") {
```
chraibi, line 241: Maybe remove the option `format` from the inifile?

libcore/src/IO/IniFileParser.cpp
```diff
@@ -283,13 +283,6 @@ bool IniFileParser::ParseHeader(TiXmlNode * xHeader)
...
         // change it to correct one
         switch(_config->GetFileFormat()) {
-            case FileFormat::XML: {
-                if(extension != ".xml") {
-                    trajectoryFile.replace_extension(".xml");
-                    LOG_WARNING("replaced output file extension with: .xml");
-                }
-                break;
-            }
             case FileFormat::TXT: {
```
chraibi, line 286: I just wanted to suggest removing this switch. But, maybe in the future we want to introduce a new format (binary)..

zeroset:
> (..and change in the nav-bar left "Input files" to "File formats")
Why should i change input files to file formats? Isn't it still the section on input files?

libcore/src/IO/IniFileParser.cpp
zeroset: I would keep this, because it also emits a warning if the format is not set or wrong.

libcore/src/IO/IniFileParser.cpp
zeroset: I removed the switch, we could add it later on again.

zeroset:
Removed Warnings and fixed documentation.

PR #738, opened 2020-04-14
Bug fix: FF router with closed transitions

schroedtert:
When in the final subroom/room the FF router also considered closed doors. Hence the always went to the closest door even if it is closed and they should be redirected.

- Closed doors are now not considered by the router when in final subroom/room
- Add simple test: One corridor with doors at each end. Closer door to pedestrian is closed. Check if open door is used. (Tested for all routers)

### DIFF
--- a/libcore/src/routing/ff_router/ffRouter.cpp
+++ b/libcore/src/routing/ff_router/ffRouter.cpp
@@ -362,8 +362,9 @@ int FFRouter::FindExit(Pedestrian * p)
     validFinalDoor.clear();
     if(goalID == -1) {
         for(auto & pairDoor : _ExitsByUID) {
-            //we add the all exits
-            if(!_building->GetTransitionByUID(pairDoor.first)->IsClose()) {
+            // we add all open/temp_close exits
+            if(_building->GetTransitionByUID(pairDoor.first)->IsOpen() ||
+               _building->GetTransitionByUID(pairDoor.first)->IsTempClose()) {
                 validFinalDoor.emplace_back(pairDoor.first); //UID
             }
         }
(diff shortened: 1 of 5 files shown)


libcore/src/routing/ff_router/ffRouter.cpp
```diff
@@ -362,8 +362,10 @@ int FFRouter::FindExit(Pedestrian * p)
     validFinalDoor.clear();
     if(goalID == -1) {
         for(auto & pairDoor : _ExitsByUID) {
-            //we add the all exits,
-            validFinalDoor.emplace_back(pairDoor.first); //UID
+            //we add the all exits
```
zeroset, line 365: We only add open exits ;)

libcore/src/routing/ff_router/ffRouter.cpp
```diff
@@ -398,7 +400,9 @@ int FFRouter::FindExit(Pedestrian * p)
         for(auto & transI : _building->GetRoom(p->GetRoomID())
                                 ->GetSubRoom(p->GetSubRoomID())
                                 ->GetAllTransitions()) {
-            DoorUIDsOfRoom.emplace_back(transI->GetUniqueID());
+            if(!transI->IsClose()) {
```
zeroset, line 403: Couldn't we use IsOpen to avoid the negation?

libcore/src/routing/ff_router/ffRouter.cpp
schroedtert: Changed

libcore/src/routing/ff_router/ffRouter.cpp
schroedtert: Unfortunately no, the door should also be considered if the door is temp_close. Maybe at some point a function which checks for open and temp_close should be introduced.

PR #740, opened 2020-04-15
Move train parser to separate file

schroedtert:
Reintroduce the train file parsing as free functions in the same matter as the event file parsing.

Added simple test for checking if the current functionalities are preserved. Test will be updated and improved with follow up PRs for trains.


libcore/src/Simulation.cpp
```diff
@@ -556,16 +550,16 @@ double Simulation::RunBody(double maxSimTime)
                 int id           = atoi(strs[1].c_str());
                 std::string type = Trans->GetCaption();
                 trainOutflow[id] += Trans->GetDoorUsage();
-                if(trainOutflow[id] >= TrainTypes[type]->nmax) {
+                if(trainOutflow[id] >= _building->GetTrainTypes().at(type).nmax) {
```
zeroset, line 553: We should catch the exception... Maybe outside this function and Log the Error.

libcore/src/Simulation.cpp
```diff
@@ -1080,22 +1096,22 @@ bool Simulation::TrainTraffic()
...
     auto now    = Pedestrian::GetGlobalTime();
-    for(auto && tab : TrainTimeTables) {
-        trainType  = tab.second->type;
-        trainId    = tab.second->id;
-        trackStart = tab.second->pstart;
-        trackEnd   = tab.second->pend;
-        if(!tab.second->arrival && (now >= tab.second->tin) && (now <= tab.second->tout)) {
-            trainHere                            = true;
-            TrainTimeTables.at(trainId)->arrival = true;
+    for(auto & [id, tab] : _building->GetTrainTimeTables()) {
```
zeroset, line 1099: const?

systemtest/train_tests/simple_train_test/runtest_simple_train.py
```diff
@@ -0,0 +1,35 @@
...
+
+def runtest(inifile, trajfile):
+    fps, N, traj = parse_file(trajfile)
+
+    maxFrame = np.max(traj[:, 1])
+    time = maxFrame / fps
+
+    if time > 50:
+        logging.critical("%s exists with failure!" % argv[0])
+        logging.critical("simulation took longer than 44s")
```
zeroset, line 27:
```suggestion
        logging.critical("simulation took longer than 50s")
```

libcore/src/Simulation.cpp
schroedtert: changed

libcore/src/Simulation.cpp
schroedtert: added try catch around RunStandardSimulation in main.cpp

PR #746, opened 2020-04-20
Trains in event manager

schroedtert:
Extend the existing event mechanism to handle train events:  
- Adds an additional abstraction level to events. Events is now an interface for any geometry changing, user defined actions.
- Events regarding the doors are moved to class `DoorEvent`
- Moves the logic for arriving and departing trains from Simulation and Building to class `TrainEvent`
- Updates the parsing of DoorEvents and TrainEvents


libcore/src/events/EventManager.cpp
```diff
@@ -38,74 +38,42 @@ EventManager::EventManager(Building * _b)
     _building = _b;
 }
 
-void EventManager::AddEvents(std::vector<Event> events)
+void EventManager::AddEvent(std::unique_ptr<Event> event)
```
zeroset, line 41: Could you use R-value references here to avoid the erroneous use with a unique pointer. With r value ref it is clear, that the memory is claimed in the function.

libcore/src/events/EventManager.cpp
```diff
@@ -38,74 +38,42 @@ EventManager::EventManager(Building * _b)
...
 {
-    _events.insert(_events.end(), events.begin(), events.end());
+    event->SetBuilding(_building);
+    _events.emplace_back(std::move(event));
+    std::sort(std::begin(_events), std::end(_events), [](auto & event1, auto & event2) {
```
zeroset, line 45: Couldn't we use a sorted data structure and insert the element at the correct position instead of resorting every time?

libcore/src/events/EventManager.cpp
```diff
@@ -38,74 +38,42 @@ EventManager::EventManager(Building * _b)
...
-                                "temp_close. Default: do nothing");
-                    break;
-            }
-            eventProcessed = true;
-        }
-    }
-    return eventProcessed;
-}
+    auto firstEventAtTime = std::upper_bound(
+        std::begin(_events), std::end(_events), eventMin, [](auto & val, auto & event) {
```
zeroset, line 64: Maybe we should implement a lower operator on Event? Would clean up this part, but maybe it is not clear what lower means on event?

libcore/src/events/TrainEvent.cpp
```diff
@@ -0,0 +1,320 @@
+#include "TrainEvent.h"
+
+#include "geometry/SubRoom.h"
+#include "geometry/Wall.h"
+
+void TrainEvent::Process()
```
zeroset, line 6: I think the event handler should be implemented separately. I would like to have a very small event object and an additional event handler object. When an event is triggered, the corresponding event handler should be called for this event. This also avoids that the event objects are bloated. Maybe we should do this in another PR/Issue. This would be too much and not focused on this PR.

libcore/src/events/TrainEvent.cpp
```diff
@@ -0,0 +1,320 @@
...
+            break;
+        case EventAction::TRAIN_DEPARTURE:
+            TrainDeparture();
+            break;
+        default:
+            throw std::runtime_error("Wrong EventAction for TrainEvent!");
+    }
+}
+
+std::string TrainEvent::GetDescription() const
```
zeroset, line 21: I would like to have this as template function having the EventAction as Non-Type Template parameter.

libcore/src/events/EventManager.cpp
schroedtert: After discussion: no change. From the cpp core guidelines: R.32: Take a unique_ptr<widget> parameter to express that a function assumes ownership of a widget

libcore/src/events/EventManager.cpp
schroedtert: After discussion: no changes. At the moment is the number of events not significant to raise performance issues.

libcore/src/events/EventManager.cpp
schroedtert: Unfortunately comparison operators won't work as unique_ptr are saved in the container, which would lead to comparison of the pointers and not the content.

libcore/src/events/TrainEvent.cpp
schroedtert: Created issue: #764

libcore/src/events/TrainEvent.cpp
schroedtert: see #764

Issue #747, opened 2020-04-21
Inconsistent definition of points for geometry and measurement area


JetteSchumann:
When testing Method D I had an issue with the definition of the mearuement area.

I defined walls in the geometry file as follows:
```xml
<polygon caption="wall">
<vertex px="10.0" py="0.0"/>
<vertex px="-0.5" py="0.0"/>
<vertex px="-0.5" py="6.0"/>
<vertex px="10.0" py="6.0"/>
</polygon>
```

Based on that I created a measurement area in the jpsreport ini file:
```xml
<area_B id="1" type="BoundingBox" zPos="None">
<vertex px="4.0" py="0.0"/>
<vertex px="4.0" py="6.0"/>
<vertex px="8.0" py="6.0"/>
<vertex px="8.0" py="0.0"/>
</area_B>
```

After running jpsreport the logging says "INFO:	Success with Method D using measurement area id 1!" - which is not correct. All output data are nan.

Thanks to @gjaeger I found out that the measurement area defintion must be:
```xml
<area_B id="1" type="BoundingBox" zPos="None">
<vertex x="4.0" y="0.0"/>
<vertex x="4.0" y="6.0"/>
<vertex x="8.0" y="6.0"/>
<vertex x="8.0" y="0.0"/>
</area_B>
```

This is quite irritating. We should have consistent naming here and the input error is not detected. This problem should occur whenever a (faulty) box-mearsurement area is used, not Method D only.

Does anyone know where the "p" comes from? Is it standing for "point"?
Should we use `px` when defining `area_B` and `area_L`?

gjaeger:
> Does anyone know where the "p" comes from? Is it standing for "point"?

"p" stands for "point".

Issue #692, opened 2020-02-13
Test Method D


chraibi:
Test Method D against simple scenarios.

Create (with `jpscore`) simple simulations, where the outcome of the method should be straightforward and easy to follow and reproduce.

PR #753, opened 2020-04-24
Test method D

JetteSchumann:
Added test for validating the calculation of the classical voronoi data in order to solve #692 

Next steps:
- test of `cut_off` parameter
- test IFD output
- change MA definition after solving #747


systemtest/jpsreport_tests/method_D/D03_corridor_medium_density/D03_corridor_medium_density.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+# ---------
+# Test of classical voronoi output
+# Pedestrians are moving in a 6x6 grid in x-direction with v=1 m/s
+# They have a fixed horizontal and vertical distance of 1.0m
+# Due to the alignment the inner voronoi cells should be squares
+# The inner voronoi cells should lie entirely in the specified square measurement area for frame 109
```
chraibi, line 7: lay

zeroset:
Can you write the test function in a utility file to avoid copy and pasting the test between low, medium and high resolution tests?

JetteSchumann:
Added tests for cut off option and IFD values.
Next steps:
- add sketches of the scnearios/ testing idea
- implement tests for voronoi cells intersecting with geometry
- still need to change MA definition

JetteSchumann:
### Documentation of the test scenarios

Basic idea is to create a grid of trajectories with pedestrians having a constant vertical and horizontal distance to each other. The voronoi cells of the inner pedestrians (blue area) are supposed to be squares and should not change over time. A measurement area (MA) is constructed at a certain position so that for a specified frame the voronoi cells of the inner pedestrians are entirely included (and only those). The averaged density for the MA should be 1/(area of a square).

Voronoi cells of the outer pedestrians are not validated since they are effected by the geometry or `cut_by_circle`option.

[setup_basic.pdf](https://github.com/JuPedSim/jpscore/files/4550919/setup_basic.pdf)

JetteSchumann:
@gjaeger and @chraibi any further tests needed?
not sure about the `profiles`option...

systemtest/jpsreport_tests/CMakeLists.txt
```diff
@@ -17,7 +26,16 @@ set_tests_properties(
   05_method_A
   01_method_B
   01_method_C
-  01_method_D
+        D01_test_reference_data
```
zeroset, line 29: please use the same indention

JetteSchumann: Done. Happened when automatically reformating the code...

PR #759, opened 2020-04-30
Consistent defintion of points in ini and geometry files

JetteSchumann:
Added additional `px` and `py`attritbutes to `area_L` and `area_B`.
Lots of redundant code in `ArgumentParser.cpp` since the old definition with `x` and `y` is still supported.

jpsreport tests must be adapted afterwards.

Once the old definition is not supported anymore the attributes of `area_L` should be changed back to `required` because the error when these attributes are missing is not caught at this point.
Missing attributes for `area_B` lead to creation of a big bounding box which should be fine.

Solves #747

### DIFF
--- a/jpsreport/general/ArgumentParser.cpp
+++ b/jpsreport/general/ArgumentParser.cpp
@@ -419,7 +419,7 @@ bool ArgumentParser::ParseIniFile(const string & inifile)
             for(TiXmlElement * xVertex = xMeasurementArea_B->FirstChildElement("vertex"); xVertex;
                 xVertex                = xVertex->NextSiblingElement("vertex")) {
                 // Note: Attributes are optional, their existence needs to be checked since xmltof returns 0.0 if unknown
-                if(xVertex->Attribute("x") != NULL and xVertex->Attribute("y") != NULL) {
+                if(xVertex->Attribute("x") != NULL && xVertex->Attribute("y") != NULL) {
                     // DEPRECATED FORMAT should be removed in future
                     double box_px = xmltof(xVertex->Attribute("x")) * M2CM;
                     double box_py = xmltof(xVertex->Attribute("y")) * M2CM;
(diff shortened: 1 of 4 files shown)


gjaeger:
@JetteSchumann 

> Once the old definition is not supported anymore the attributes of area_L should be changed back to required because the error when these attributes are missing is not caught at this point.
> Missing attributes for area_B lead to creation of a big bounding box which should be fine.

How can we keep this requirement for later?

gjaeger:
@JetteSchumann 
Ok, I found the note in the source code. 👍

jpsreport/general/ArgumentParser.cpp
```diff
@@ -418,12 +418,26 @@ bool ArgumentParser::ParseIniFile(const string & inifile)
...
             for(TiXmlElement * xVertex = xMeasurementArea_B->FirstChildElement("vertex"); xVertex;
                 xVertex                = xVertex->NextSiblingElement("vertex")) {
-                double box_px = xmltof(xVertex->Attribute("x")) * M2CM;
-                double box_py = xmltof(xVertex->Attribute("y")) * M2CM;
-                boost::geometry::append(poly, boost::geometry::make<point_2d>(box_px, box_py));
-                Log->Write(
-                    "\t\tMeasure area points  < %.3f, %.3f>", box_px * CMtoM, box_py * CMtoM);
-                num_verteces++;
+                // Note: Attributes are optional, their existence needs to be checked since xmltof returns 0.0 if unknown
+                if(xVertex->Attribute("x") != NULL and xVertex->Attribute("y") != NULL) {
```
schroedtert: Better check against `nullptr` to explicit state what is checked here. `xVertex->Attribute("foo")` returns a `char *` but `NULL` can be either 
an integer literal with value zero or prvalue of type std::nullptr_t.

A more detailed explanation can be found here: https://cpp4arduino.com/2018/10/26/why-cpp-programmers-dont-use-null.html

JetteSchumann: Thanks! Done.

Issue #760, opened 2020-05-03
missing seed value


chraibi:
From the  [documentation](http://www.jupedsim.org/jpscore_inifile.html#header)

If there is no seed tag in the inifile then the seed is default 0. If an empty seed tag is used the seed is randomly initialized. 

That means 

```
<seed></seed>
```

However, we might expect that a missing seed-tag leads to random initial conditions as well. 

ACs: 
- Random seed should be random initialized if no seed is provided. 
- Check & update existing Documentation.

PR #765, opened 2020-05-06
Fix missing seed value (#760)

xuqiancheng:
Now a missing seed-tag leads to random initial conditions as well.

Existing documentation is updated.


libcore/src/IO/IniFileParser.cpp
```diff
@@ -174,6 +174,8 @@ bool IniFileParser::ParseHeader(TiXmlNode * xHeader)
         } else {
             _config->SetSeed((unsigned int) time(NULL));
         }
+    } else {
+        _config->SetSeed((unsigned int) time(NULL));
```
zeroset, line 178: Could you pls use static_cast<unsigned int>  instead of c style cast. Pls also fix this two lines above. Thanks

xuqiancheng: see commit 4c3633e

Issue #742, opened 2020-04-20
Train: Add Doors to Geometry 


zeroset:
For arriving Trains we must add new Doors to the geometry. 

Non-Goals:
- Fix/Correct Geometry

schroedtert:
Closed with #771

Issue #745, opened 2020-04-20
Train: Train leaving the station


zeroset:
After a train left the station the old walls must be added again.

schroedtert:
Done with #781

PR #771, opened 2020-05-11
Move logic for adding trains to geometry::helper

schroedtert:
Instead of having the logic for computing the projected door coordinates onto the track walls and the splitting of the track walls in the TrainEvent the functionality has been moved to the geometry helper namespace. 

- ComputeTrainDoorCoordinates: function for computing the projected door coordinates onto the trackwalls
- AddTrainDoors: Interface for the train event to update the subroom, delete or add some walls and add the train doors
- SplitWalls: Split the walls on the given cutting points, returns walls to add/remove

Additionally the container for saving temporarily deleted/added walls, doors for trains have been made private for the Building, and corresponding getter/setter have been added.

WIP: 
- [x] improved unittests for the SplitWalls method   
- [x] missing documentation

Solves #742


libcore/src/events/TrainEvent.cpp
```diff
@@ -240,80 +89,89 @@ void TrainEvent::TrainDeparture()
...
-        auto wall = *it;
-        if(it != tempWalls.end()) {
-            tempWalls.erase(it);
+    auto tempWalls = _building->GetTrainAddedWalls(_trainID);
+    if(tempWalls.has_value()) {
+        for(auto it = tempWalls.value().begin(); it != tempWalls.value().end();) {
+            auto wall = *it;
+            if(it != tempWalls.value().end()) {
+                tempWalls.value().erase(it);
+            }
```
zeroset, line 98: This seems to be quite dangerous, as i am not sure if the iterator is still valid afterwards. Will this be changed in following PRs?

libcore/src/geometry/Building.cpp
```diff
@@ -1252,4 +1187,100 @@ std::vector<TrainType> Building::GetTrainTypes()
...
+        _trainAddedWalls.emplace(trainID, std::vector<Wall>{trainAddedWall});
+    }
+}
+
+void Building::SetTrainAddedWalls(int trainID, std::vector<Wall> trainAddedWalls)
+{
+    _trainAddedWalls[trainID] = trainAddedWalls;
+}
+
+std::optional<std::vector<Wall>> Building::GetTrainAddedWalls(int trainID)
```
zeroset, line 1206: Can we name this `GetTrainWallsAdded` to sound more like the walls added from the train? Or do i misunderstand?

libcore/src/events/TrainEvent.cpp
schroedtert: It is, will be tackled soon when solving #745

libcore/src/geometry/Building.cpp
schroedtert: Changed, also changed the names of the other functions to fit the naming

PR #777, opened 2020-05-15
Modifying Utest Methods

### LINKED ISSUE #775: jpsreport, Modifying Utest Methods
Extension of existing test _ini_Methods.xml_

gjaeger:
Extension of the utest by variation of the variables `cut_of_circle` and `radius` for methods D, I and J. This pull request contains the necessary ini- and py-files and reference data for all methods. The integration into the test suite did not take place. fix #775


zeroset:
Thanks for the input, I would prefer having this as a systemtest running in CI on Teamcity. Therefor we need a python script running jpsreport using the inifiles provided. Additionally I would prefer just diffing the results file to be exact the same instead of testing them using Kolmogorow Smirnow test.

gjaeger:
I have marked this PR as a _Draft_ because I am still working on it. ;-)

gjaeger:
> Additionally I would prefer just diffing the results file to be exact the same instead of testing them using Kolmogorow Smirnow test.

With commit 42527c9 I would like to make a proposal for discussion:

If we only test if the results are exactly the same, then we can get a wrong result with different precision. The numpy-function [`array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html) only returns `True` if the results are exactly the same. The [Kolmogorov-Smirnov statistic test](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.ks_2samp.html#scipy.stats.ks_2samp) allows a statement about the deviation. The tolerance threshold must then be defined. Alternatively there is the numpy function [`allclose`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html).

@zeroset The file `runtest_Method_2_D.py`contains a proposal for the modification.

jpsreport/Utest/Methods/runtest_Method_A.py
```diff
@@ -1,53 +0,0 @@
-#!/usr/bin/env python
```
zeroset, line 1: why did you delete these files? Or did you just move them?

systemtest/jpsreport_tests/CMakeLists.txt
```diff
@@ -34,6 +34,15 @@ add_test(NAME J08_corridor_low_cut_off COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURR
...
 add_test(NAME J10_corridor_cut_off_velocity COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/method_J/J10_corridor_cut_off_velocity/J10_corridor_cut_off_velocity.py ${jpsreport_exe})
 
+add_test(NAME reference_cut_off_disabled COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/reference_tests/cut_off_disabled/test_cut_off_disabled.py ${jpsreport_exe})
+add_test(NAME reference_cut_off_edges COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/reference_tests/cut_off_edges/test_cut_off_edges.py ${jpsreport_exe})
+add_test(NAME reference_cut_off_enabled COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/reference_tests/cut_off_enabled/test_cut_off_enabled.py ${jpsreport_exe})
+add_test(NAME reference_cut_off_large COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/reference_tests/cut_off_large/test_cut_off_large.py ${jpsreport_exe})
+add_test(NAME reference_cut_off_small COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/reference_tests/cut_off_small/test_cut_off_small.py ${jpsreport_exe})
+add_test(NAME reference_methods_ABC COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/reference_tests/methods_ABC/test_methods_ABC.py ${jpsreport_exe})
```
zeroset, line 42: Didn't we just write reference tests for Method D, I and J? Do we have some for ABC?

jpsreport/Utest/Methods/runtest_Method_A.py
JetteSchumann: The original tests created one tests for each outputfile to validated. I restructured this so that all generated outputfiles (based on one inifile) are validated/compared with a reference directory in one test.
Also, statistics functions have been used to check/assess the detected difference in outputfiles. I removed this.
I deleted the old tests as there were major changed. Should we keep the old files?

systemtest/jpsreport_tests/CMakeLists.txt
JetteSchumann: The original inifiles executed Method A,B,C as well. I separated this from testing of D,I,J

PR #784, opened 2020-05-20
Change definition of isLineSegment

### LINKED ISSUE #762: Bug: IsLineSegment depends on line length
It seems to be an accuracy issue when checking if a point depends to a line segment. 

Point (4.00036, 0)

Test 1:
Line (4, -1.5) - (4, 1.5) -> Point is not in line segment, as they are not colinear (`std::abs(crossProduct) > J_EPS` see below)

Test 2:
Line (4, -0.4) - (4, -0.4) -> Point is in line segment

```cpp
// From Macros.h
#define J_EPS 0.001        //0.001

// From Line.cpp
/*
 *  Prüft, ob Punkt p im Liniensegment enthalten ist
 * algorithm from:
 * http://stackoverflow.com/questions/328107/how-can-you-determine-a-point-is-between-two-other-points-on-a-line-segment
 *
 */
bool Line::IsInLineSegment(const Point & p) const
{
    Point differenceTwoAndOne = _point2 - _point1;
    Point differencePAndOne   = p - _point1;

    // cross product to check if point i colinear
    auto crossProduct = differenceTwoAndOne.CrossProduct(differencePAndOne);
    if(std::abs(crossProduct) > J_EPS)
        return false;

    // dotproduct and distSquared to check if point is in segment and not just in line
    double dotp = differencePAndOne.ScalarProduct(differenceTwoAndOne);

    if(dotp < 0.) {
        return false;
    }

    if(dotp > (differenceTwoAndOne).NormSquare()) {
        return false;
    }
    return true;
}
```

Suggestion for discussion:
- decrease J_EPS
- use `boost::geometry::distance()` with a threshold
- use 'boost::geometry::within()` with a certain threshold polygon
- use other algorithm

chraibi:
There are two functions in the class LIne:

1. NearlyInLineSegment()
2. isInLineSegment()

The only difference between these functions is precision.
While in the first (used for correct geometry)
we check if a point is somewhere near the line, in the second
to a certain limited precision "in" the segment line.

In this PR the first function is removed and the second is changed such that the precision is an input argument. In this way, one function can serve both purposes.

fixes #762


libcore/test/catch2/geometry/LineTest.cpp
```diff
@@ -231,6 +231,20 @@ TEST_CASE("geometry/Line/IsInLineSegment", "[geometry][Line][IsInLineSegment]")
...
+        Line shortLine{{4., -0.4}, {4., 0.4}};
+        bool isInShort = shortLine.IsInLineSegment(point);
+
+        Line longLine{{4., -1.95}, {4., 1.95}};
+        bool isInLong = longLine.IsInLineSegment(point);
+
+        REQUIRE(isInShort);
+        REQUIRE(isInLong);
+    }
+
```
schroedtert, line 247: Could you rewrite the test, for multiple of this cases? Lines varying in length, but the distance from the line stays the same. So should be the result.

libcore/src/geometry/Point.cpp
```diff
@@ -220,7 +220,7 @@ const Point operator/(const Point & p, double f)
         return Point(p._x / f, p._y / f);
     else {
         std::cout << "Warning: Point::/operator. dividand " << f
-                  << " is to small. Set it to 1 instead" << std::endl;
+                  << " is too small. Set it to 1 instead" << std::endl;
```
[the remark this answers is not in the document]
chraibi: okay!

libcore/test/catch2/geometry/LineTest.cpp
chraibi: done.

libcore/src/geometry/Line.cpp
```diff
@@ -219,31 +219,11 @@ Point Line::ShortestPoint(const Point & p) const
...
-
-    // dotproduct and distSquared to check if point is in segment and not just in line
-    double dotp = differencePAndOne.ScalarProduct(differenceTwoAndOne);
-
-    if(dotp < 0.) {
-        return false;
-    }
-
-    return dotp <= (differenceTwoAndOne).NormSquare();
+    return fabs((_point1 - p).Norm() + (_point2 - p).Norm() - (_point2 - _point1).Norm()) < 1E-6;
```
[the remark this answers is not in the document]
chraibi: removed `NearlyInLineSegment` and changes the signature of `IsInLineSegment`

libcore/src/geometry/Line.cpp
```diff
@@ -218,40 +218,13 @@ Point Line::ShortestPoint(const Point & p) const
...
-
-    return dotp <= (differenceTwoAndOne).NormSquare();
-}
-
-bool Line::NearlyInLineSegment(const Point & p) const
+bool Line::IsInLineSegment(const Point & p, double tolerance) const
 {
     return fabs((_point1 - p).Norm() + (_point2 - p).Norm() - (_point2 - _point1).Norm()) <
-           J_EPS_DIST; // old version
+           tolerance;
```
schroedtert, line 224: If my computations are correct, this still depends on the length of the line.
Example: 
Point p: (0.1, 0.1)
Line line1: (0, 0) - (5,0)
fabs((_point1 - p).Norm() + (_point2 - p).Norm() - (_point2 - _point1).Norm()) = 0.042

Line line2: (0, 0) - (1,0)
fabs((_point1 - p).Norm() + (_point2 - p).Norm() - (_point2 - _point1).Norm()) = 0.047

schroedtert, line 224: If my computations are correct, this still depends on the length of the line.
Example:
Point p: (0.1, 0.1)
Line line1: (0, 0) - (5,0)
fabs((_point1 - p).Norm() + (_point2 - p).Norm() - (_point2 - _point1).Norm()) = 0.042

Line line2: (0, 0) - (1,0)
fabs((_point1 - p).Norm() + (_point2 - p).Norm() - (_point2 - _point1).Norm()) = 0.047

zeroset:

PR #787, opened 2020-05-24
Exit jpsreport, if measurement are is misconfigured.

### LINKED ISSUE #706: jpsreport: Method D with a measurement method out of bounds
What happens actually if the measurement method is not in the geometry?

`jpsreport` gives the following error: 

```
ERROR: 	 No polygon containing the measurement id [1]
```

and goes on by producing a ton of warnings

```
Warning:	Pedestrian at <x=-4.0851, y=4.1511> is not in geometry and not considered in analysis!
```

then ends with an optimistic message 

```
INFO:	Success with Method D using measurement area id 1!
```

In this case this [if](https://github.com/JuPedSim/jpscore/blob/master/jpsreport/Analysis.cpp#L234) is not entered, so that `geoPoly[area->_id]` does not get initialised.

How do you suggest to deal with this situation?

******

JuPedSim - JPSreport

Current date   : Fri Feb 21 14:47:21 2020
Version        : 0.8.4
Compiler       : clang++ (11.0.0)
Commit hash    : v0.8.4-923-geaf22278
Commit date    : Thu Feb 20 19:18:20 2020
Branch         : master

chraibi:
There is not a sane case where this situation should be tolerated.
Therefore, exit calculation at once. The user is expected to
configure measurement areas properly.

Fixes  #706

### DIFF
--- a/jpsreport/Analysis.cpp
+++ b/jpsreport/Analysis.cpp
@@ -44,9 +44,11 @@
 #include <Logger.h>
 #include <algorithm> // std::min_element, std::max_element
 #include <cfloat>
+#include <fmt/format.h>
 #include <fstream>
 #include <iostream>
 #include <sstream>
+#include <stdexcept>
 #include <stdlib.h>
 #include <vector>
 
@@ -246,8 +248,9 @@ std::map<int, polygon_2d> Analysis::ReadGeometry(
         } //room
 
         if(geoPoly.count(area->_id) == 0) {
-            LOG_ERROR("No polygon containing the measurement id {}.", area->_id);
             geoPoly[area->_id] = area->_poly;
+            throw std::runtime_error(
+                fmt::format(FMT_STRING("No polygon containing the measurement id {}."), area->_id));
         }
     }
 
--- a/jpsreport/main.cpp
+++ b/jpsreport/main.cpp
@@ -35,6 +35,7 @@
 
 #include <Logger.h>
 #include <chrono>
+#include <exception>
 
 using namespace std;
 using namespace std::chrono;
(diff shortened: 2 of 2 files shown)


jpsreport/Analysis.cpp
```diff
@@ -264,6 +264,7 @@ std::map<int, polygon_2d> Analysis::ReadGeometry(
 
         if(geoPoly.count(area->_id) == 0) {
             Log->Write("ERROR: \t No polygon containing the measurement id [%d]", area->_id);
+            exit(EXIT_FAILURE);
```
schroedtert, line 267: I suggest that we do not use exit not during the run of the program but instead throw exceptions. These exception should be caught on top level. Similar to jpscore (see main.cpp)

schroedtert, line 267: There should be no code after a exit/exception as will never be called

jpsreport/methods/Method_D.cpp
```diff
@@ -131,11 +131,6 @@ bool Method_D::Process(
         //------------------------------Remove peds outside geometry------------------------------------------
         for(int i = 0; i < (int) IdInFrame.size(); i++) {
             if(false == within(point_2d(round(XInFrame[i]), round(YInFrame[i])), _geoPoly)) {
-                Log->Write(
-                    "Warning:\tPedestrian at <x=%.4f, y=%.4f> is not in geometry and not "
-                    "considered in analysis!",
-                    XInFrame[i] * CMtoM,
-                    YInFrame[i] * CMtoM);
```
zeroset, line 138: Why did you remove this?

chraibi: because sometimes when the geometry does not fit the trajectories, there are tones of warnings and that's very noisy.

PR #774, opened 2020-05-14
Separate attributes for Method D, I and J

JetteSchumann:
Method options for D, I and J need to be read in and handled separately. Otherwise when Method D, I and J are defined in the same jpsrport ini-file their options are overwritten and the same configuration is used for all of them.

The ArgumentParser and Analysis must be adapted. 

Solves #756

Issue #756, opened 2020-04-29
Overwritten method option


JetteSchumann:
When enabling several methods in the jpsreport inifile, the `cut_by_circle` option of the last defined method is taken for all of them. So it is `true` for method D, I and J - but should be `false` for method D and I for the following definition:

```xml
  <method_D enabled="true">
    <measurement_area id="1" start_frame="None" stop_frame="None" get_individual_FD="true"/>
    <one_dimensional enabled="false"/>
    <cut_by_circle enabled="false" radius="1.0" edges="10"/>
  </method_D>

  <method_I enabled="true">
    <measurement_area id="2" start_frame="None" stop_frame="None" get_individual_FD="true"/>
    <one_dimensional enabled="false"/>
    <cut_by_circle enabled="false" radius="1.0" edges="10"/>
  </method_I>

  <method_J enabled="true">
    <measurement_area id="1" start_frame="None" stop_frame="None" get_individual_FD="true"/>
    <one_dimensional enabled="false"/>
    <cut_by_circle enabled="true" radius="1.0" edges="10"/>
  </method_J>
```

Seems that the option is overwritten. Setting of other paramters should be checked as well.

JetteSchumann:
I checked the code. The argument parser cannot handle all parameters for method D, I and J separatly. That applies to these parameter:

```c++
    _isCutByCircle          = false;
    _cutRadius              = 50;
    _circleEdges            = 6;
    _isOneDimensional       = false;
    _isGetProfile           = false;
    _steadyStart            = 100;
    _steadyEnd              = 1000;
    _grid_size_X            = 10;
    _grid_size_Y            = 10;
    std::vector<bool> _individual_FD_flags;
```

I will add parameteres for each method as it is done for e.g. 

```c++
    std::vector<int> _start_frames_MethodD;
    std::vector<int> _stop_frames_MethodD;
    std::vector<int> _start_frames_MethodI;
    std::vector<int> _stop_frames_MethodI;
    std::vector<int> _start_frames_MethodJ;
    std::vector<int> _stop_frames_MethodJ;
```

JetteSchumann:
Found out that steady state paramerters are read in for method D, I and J but they are never used.

```c++
    _steadyStart            = 100;
    _steadyEnd              = 1000;
```

Any idea why they were needed once?
If they are not used anymore I would suggest to delete them in the context of this issue.

JetteSchumann:
Issue have been solved by PR #774 .
[...]

PR #794, opened 2020-05-26
Improved Logging for Method D, I and J

JetteSchumann:
Moved logging information of the chosen method to the beginning before options are parsed.
Logging of the chosen options per method follows.

Solves #756 after restructuring data handling with PR #774


zeroset:
Shouldn't we change this to the spdlog based logging? Or are we doing this after libshared?

JetteSchumann:
thought we would wait... wasn't this PR blocked before?

Issue #616, opened 2019-12-16
Clean up pedestrian class


zeroset:
Main goal: Reduce class size and move variables to the corresponding classes where they belong and remove stuff which is not needed.

If you have any objections, please comment here.

Steps:
- [ ] mark stuff that should either be removed or moved to different class + discussion
- [ ] remove not implemented not called functions, not used variables
- [ ] move variables to responsible classes
- [ ] check if functions are returning what their function name suggests

PR #812, opened 2020-07-07
Clean up pedestrian class

schroedtert:
Removes following from the pedestrian class:
- not implemented
- not called functions
- not used variables

Minor fixes:
- Fixes most clang-tidy warnings
- renaming of variables to fit code convention (not raising clang-tidy warnings anymore) 

Subtask of #616


libcore/src/pedestrian/Pedestrian.cpp
```diff
@@ -1101,16 +815,17 @@ int Pedestrian::GetColor() const
             double v0 = GetV0Norm();
             if(v0 != 0.0) {
                 double v = GetV().Norm();
-                color    = (int) (v / v0 * 255);
+                color    = static_cast<int>(v / v0 * 255);
             }
             return color;
-        } break;
```
JetteSchumann, line 1107: you have removed a `break` here. this didn't change the logical part?

libcore/src/pedestrian/Pedestrian.cpp
```diff
@@ -1123,7 +838,7 @@ int Pedestrian::GetColor() const
                                           // colors clearly distinguishable form
                                           // each other
             return (colors[_group % colors.size()]);
-        } break;
```
JetteSchumann, line 1126: same here

libcore/src/pedestrian/Pedestrian.cpp
schroedtert: no, as it is unreachable code. Due to the `return` before the function ends before the `break` can be reached.

libcore/src/pedestrian/Pedestrian.cpp
schroedtert: see above

PR #817, opened 2020-07-23
documentation, add option to write logging information to log file

### LINKED ISSUE #644: Option to write logging information to log file
issue #459 _Setting a log file location removes errors from output to stdout_ was closed because with the logging rework no log file can be set anymore.

The possibility to write the logging-information into a (log-)file would be good. Can we implement this option again?

gjaeger:
Added notes on logfiles when calling the program jpscore and jpsreport to write logging information to log file.

resolve #644

### DIFF
--- a/docs/pages/jpscore/jpscore_introduction.md
+++ b/docs/pages/jpscore/jpscore_introduction.md
@@ -48,10 +48,10 @@ Taking the 7th demo as input, we run a simulation as follows:
  ./bin/jpscore  demos/corner_ini.xml
 ```
 
-It is possible to redirect the output and write it to a file like:
+Information on processing steps, warnings and errors of jpscore are written to the console. This output can be redirected to a file by:
 
 ```bash
- ./bin/jpscore  demos/corner_ini.xml &> log_example.txt
+ ./bin/jpscore  demos/corner_ini.xml > log_example.txt
 ```
 
 This program call produces a trajectory file in the folder `results` in the same directory. This can be visualized with `jpsvis`
--- a/docs/pages/jpscore/jpscore_introduction.md
+++ b/docs/pages/jpscore/jpscore_introduction.md
@@ -48,7 +48,13 @@ Taking the 7th demo as input, we run a simulation as follows:
  ./bin/jpscore  demos/corner_ini.xml
 ```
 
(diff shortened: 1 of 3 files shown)


docs/pages/jpscore/jpscore_introduction.md
```diff
@@ -48,7 +48,13 @@ Taking the 7th demo as input, we run a simulation as follows:
...
 ```
 
-which produces a trajectory file in the folder `results` in the same directory. This can be visualized with `jpsvis`
+It is possible to redirect the output and write it to a file like:
+
+```bash
+ ./bin/jpscore  demos/corner_ini.xml &> log_example.txt
+```
+
+This program call produces a trajectory file in the folder `results` in the same directory. This can be visualized with `jpsvis`
```
JetteSchumann, line 57: I would keep the old text and add the logging information at the end.
My suggestion:
Information on processing steps, warnings and errors of `jpscore` are written to the console. This output can be redirected to a file by:

```bash
 ./bin/jpscore  demos/corner_ini.xml > log_example.txt
```

JetteSchumann: I found out that "&>" is not working on Windows but ">" does. Could you please change this in the jpsreport inifile documentation as well?

docs/pages/jpsreport/jpsreport_introdcution.md
```diff
@@ -39,6 +39,7 @@ run `jpsreport` in a terminal as follows:
 ./bin/jpsreport inifile.xml
 ```
 
+{%include note.html content="It is possible to redirect the output and write it to a file. See [inifile - Logfile](jpsreport_inifile#logfile)"%}
```
JetteSchumann, line 42: I think "output" is misleading. "console output" might be better.

docs/pages/jpscore/jpscore_introduction.md
gjaeger: good suggestion. Adopted

docs/pages/jpsreport/jpsreport_introdcution.md
gjaeger: good suggestion. Adopted

PR #819, opened 2020-07-27
Remove unused classes

schroedtert:
Remove unneeded file from Repo


schroedtert:
Found some more files. Will add them too, sorry.

PR #820, opened 2020-07-27
Replace std::cout/cerr with Logger calls

schroedtert:
Replace std::cout/std::cerr calls with logger in our code:
- replace by Logger call when it makes sense
- remove output otherwise

Note: Some third party libraries may still put there output to std::cout/cerr
Note: Some calls just replaced by stringstreams as the return value of some function could not be used by our Logger without more work, e.g. dsv output in jpsreport.

### DIFF
--- a/jpsreport/methods/Method_J.cpp
+++ b/jpsreport/methods/Method_J.cpp
@@ -195,7 +195,7 @@ bool Method_J::Process(
                 } else {
                     for(int i = 0; i < (int) IdInFrame.size(); i++) {
                         LOG_WARNING(
-                            "frame={}: id={} x={:.5} y={:.5}",
+                            "Polygons could not be obtained for frame={}: id={} x={:.5} y={:.5}",
                             frameNr,
                             IdInFrame[i],
                             XInFrame[i] * CMtoM,
--- a/jpsreport/methods/Method_J.cpp
+++ b/jpsreport/methods/Method_J.cpp
@@ -194,8 +194,12 @@ bool Method_J::Process(
                     }
                 } else {
                     for(int i = 0; i < (int) IdInFrame.size(); i++) {
-                        std::cout << XInFrame[i] * CMtoM << "   " << YInFrame[i] * CMtoM << "   "
-                                  << IdInFrame[i] << "\n";
+                        LOG_WARNING(
+                            "frame={}: id={} x={:.5} y={:.5}",
+                            frameNr,
+                            IdInFrame[i],
+                            XInFrame[i] * CMtoM,
+                            YInFrame[i] * CMtoM);
                     }
                     LOG_WARNING(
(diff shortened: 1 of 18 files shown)


jpsreport/methods/Method_J.cpp
```diff
@@ -194,8 +194,12 @@ bool Method_J::Process(
                     }
                 } else {
                     for(int i = 0; i < (int) IdInFrame.size(); i++) {
-                        std::cout << XInFrame[i] * CMtoM << "   " << YInFrame[i] * CMtoM << "   "
-                                  << IdInFrame[i] << "\n";
+                        LOG_WARNING(
```
JetteSchumann, line 197: I suggest to add text here: "Polygons could not be obtained for..."

libcore/src/geometry/Building.cpp
```diff
@@ -859,14 +859,14 @@ const std::vector<Pedestrian *> & Building::GetAllPedestrians() const
...
-        Pedestrian * ped1 = _allPedestrians[p];
-        if(ped->GetID() == ped1->GetID()) {
-            std::cout << "Pedestrian " << ped->GetID() << " already in the room." << std::endl;
-            return;
-        }
+    if(std::find_if(
+           std::begin(_allPedestrians), std::end(_allPedestrians), [ped](const Pedestrian * other) {
+               return ped->GetID() == other->GetID();
+           }) != std::end(_allPedestrians)) {
+        LOG_ERROR("Pedestrian {} already in the romm.", ped->GetID());
```
JetteSchumann, line 866: might be a warning only?

libcore/src/math/VelocityModel.cpp
```diff
@@ -99,7 +98,7 @@ bool VelocityModel::Init(Building * building)
         if(ped->GetExitLine())
             target = ped->GetExitLine()->ShortestPoint(ped->GetPos());
         else {
-            std::cout << "Ped " << ped->GetID() << " has no exit line in INIT\n";
+            LOG_ERROR("Ped {} has no exit line in INIT", ped->GetID());
```
JetteSchumann, line 101: warning only?

libcore/src/geometry/Building.cpp
JetteSchumann: "room"

jpsreport/methods/Method_J.cpp
schroedtert: changed

libcore/src/geometry/Building.cpp
schroedtert: changed

libcore/src/math/VelocityModel.cpp
schroedtert: changed

PR #835, opened 2020-09-07
Set working directory for reference tests

schroedtert:
At the moment the reference tests are executed in the build directory, and both tests store there results in the same folder. At my work station it happened, that the wrong results were compared to the expected output and the test failed. With this the reference tests are executed in their specific directory to avoid such issues. 
The former proposed change to the JPSTestDriver may happen later. But this should be sufficient for now.


zeroset (changes requested):

zeroset (approved):
Cool, thanks

PR #836, opened 2020-09-15
Fix missing resetting of waiting position

schroedtert:
The waiting position was not reset after the waiting of pedestrians is over, hence they returned to this position even if they were in a different room.


schroedtert:
Fixed in #862

PR #842, opened 2020-10-21
Integrate Method J in Method D

### LINKED ISSUE #809: Integrate Method J into Method D
We will extend Method D to accept different methods for the computation of the velocity v. With this Method J can be run inside Method D. The calculation of the velocity should be defined as an `std::function`. The XML parser sets the corresponding function definition as a parameter of Method D. 

- [x] Add a `std::function` parameter as velocity_calculation to Method D. 
- [x] The default velocity_calculation should be the the current calculation of Method D.
- [x] Add the functionality to exchange the velocity calculation in Method D.
- [x] Adopt tests of method I to run as Method D with corresponding parameters
- [x] Check and adopt existing inifiles in repository

JetteSchumann:
Velocity calculation method 'Voronoi' or 'Arithmetic' for Method D can be chosen now.

Velocity calculation was decoupled from density calculation and is set via a std::function parameter. Default is set to 'Voronoi' in ConfigData. Function for "Arithmetic" is defined in ArgumentParser.

Tests need to be adapted and Method J needs to be removed.

Fixes #809


gjaeger (changes requested):
Are the data for the calculation methods written in files with different file names? 

The column currently still contains "_Voronoi velocity(m/s)_". Shouldn't it be called "_Arithmetic velocity(m/s)_"? I suggest to include the calculation method in appropriate file and column names (for example: _rho_v_Voronoi_Voronoi_ or _rho_v_Voronoi_Arithemtic_).

JetteSchumann:
Decided for new File names: rho and v are followed by their calculation type
**Examples**
rho_v_Voronoi_Voronoi ... (classical output for default Method D)
IFD_rho_v_Voronoi_Arithmetic ... (IFD output with arithmetic velocity calculation)
Prf_rho_Vornoi ...
Prf_v_Voronoi ...

Most of the tests of Method J could be removed as they tested the density calculation only (which did not change). This is already done in the basic Method D tests.
New profile tests for arithmetic velocity calculation was added.

jpsreport/methods/ConfigData_D.h
```diff
@@ -47,5 +50,27 @@ struct ConfigData_DIJ {
...
+
+    std::string densityType  = "Voronoi";
+    std::string velocityType = "Voronoi";
+
+    //parameter for velocity calculation function
+    /** default:
+    Calculates the velocity of pedestrians based on their voronoi cell and their instantaneous velocity.
+    input: intersecting polygons (with MA), instantaneous velocities and measurement area.
+    output: mean velocity in the MA
+    **/
```
schroedtert, line 62: Can you make this as doxygen comment and fuse merge both?

JetteSchumann: Done

schroedtert (approved):

PR #852, opened 2020-11-26
Add frames to N_T output of Method A

gjaeger:
In the output file `Flow_NT_traj` the time is printed. Due to rounding to the second decimal place, problems may occur when calculating back to frames.


gjaeger:
@JetteSchumann The modification of the header has an error. Can you check this?

gjaeger:
@JetteSchumann Revision of the modification after our short talk.

New columns: `Frame` | `time (s)`| `cumulative pedestrians`

There is still something wrong with the modified header of the output file.

schroedtert (approved):

PR #860, opened 2021-04-30
Quick fix for 858

schroedtert:


jpsreport/methods/Method_D.cpp
```diff
@@ -156,15 +156,15 @@ bool Method_D::Process(
...
-                    ZInFrame[i] * CMtoM);
+            if(!covered_by(point_2d(round(XInFrame[i]), round(YInFrame[i])), _geoPoly)) {
+                //                LOG_WARNING(
+                //                    "Pedestrian with id <{}> at <x={:.4f}, y={:.4f}, z={:.4f}> is not in the "
+                //                    "measurement area and will not be"
+                //                    "considered in the analysis!",
+                //                    IdInFrame[i],
+                //                    XInFrame[i] * CMtoM,
+                //                    YInFrame[i] * CMtoM,
+                //                    ZInFrame[i] * CMtoM);
```
chraibi, line 167: did you just comment this warning out?

schroedtert: Yes, the warning basically says that a pedestrian is not inside a measurement area (which statically happens in most frames) and hence it is flooding the warnings, which will hide important warnings. 

Actually it should be deleted imho but in this case, the PR is just for getting the windows executables.

Ozaq:
Not part of this repo anymore.

PR #862, opened 2021-06-17
Fix faulty navigation for waiting areas

JetteSchumann:
When approaching a waiting area, formatting along a diagonal line was observed. After entering this waiting area, the pedestrians did not wait and moved on to another goal (while leaving the actual waiting area).

Screenshot 2021-06-15 at 16 13 54

Two things were changed to solve this: 
- The diagonal line (target) was narrowed down to the center
- Waiting position is reset after waiting ended

Thanks for helping @schroedtert


schroedtert (approved):
LGTM

PR #903, opened 2021-11-24
Test refactoring

JetteSchumann:
Basic draft of the new structure of pytests on the example of the reference tests. Input files are copied to a temporary directory where jpscore is executed. This makes it possible to inspect the output files without writing them to the repo.

`JPSCORE_EXECUTABLE_PATH` needs to be set when running pytest.
`tmp_path` can be set explicitly by running `pytest --basetemp=mydir`


JetteSchumann:
Some assumptions @Ozaq  and I made:

- `JPSCORE_EXECUTABLE_PATH` and `JPSCORE_SOURCE_PATH` need to be set when running pytest
- input data are always located in a directory called `input` for each test
- all files in `input` are copied to `tmp_path` where jpscore is executed
- name of the inifile is always `inifile.xml`

PR #942, opened 2022-01-31
Porting flow regulation test and refactoring

zeroset:


zeroset:
Should I remove `enable_testing` from cmake or is this used for anything else then Ctest?

systemtest/test_jpscore.py
```diff
@@ -712,16 +713,48 @@ def test_door_closes_after_max_agents(tmp_path, env):
...
+    )
+    jpscore_driver = JpsCoreDriver(
+        jpscore_path=env.jpscore_path, working_directory=tmp_path
+    )
+    jpscore_driver.run()
+
+    max_agents_dict = read_max_agents(tmp_path / "inifile.xml")
+    outflow_dict = read_outflow_from_inifile(tmp_path / "inifile.xml")
+    starting_times_dict = read_starting_times(tmp_path / "events.xml")
+    data_dict = read_flow(tmp_path)
```
Ozaq, line 758: Im not sure if I remember this correctly, but I think we discussed this and the conclusion was to not use the outflow files and instead compute the flow based on the trajectory. Esp. since we want to remove the statistics output eventually.

systemtest/driver/flow.py
```diff
@@ -80,27 +82,60 @@ def check_max_agents(flow_dict, max_agents_dict):
...
-            "Number of agents using other doors ({}) does not match number of agents in the simulation ({})".format(
-                other_door_usage, num_agents
+def read_starting_times(events_file: Path):
+    tree = ET.parse(events_file)
+    root = tree.getroot()
+    starting_times = {}
+    for event in root.iter("event"):
+        if event.attrib["state"] == "open" or event.attrib["state"] == "reset":
+            starting_times.setdefault(int(event.attrib["id"]), []).append(
+                float(event.attrib["time"])
```
Ozaq, line 92: The code in `driver` tries to offer functionality for each input file with a `.py` file of the same name, e.g. `ini file.py` `geometry.py` etc... Since this is extracted from `events.xml` I would like to see this code moved to `events.py`

systemtest/driver/flow.py
```diff
@@ -28,14 +28,27 @@ def read_flow(tmp_path: Path):
...
+    for tc in root.iter("traffic_constraints"):
+        outflow_dict = {
+            int(door.attrib["trans_id"]): float(door.attrib["outflow"])
+            for door in tc.iter("door")
+        }
+
+    assert outflow_dict, "Could not read outflow from inifile"
+    return outflow_dict
+
+
```
Ozaq, line 51: This code should be moved to `inifile.py`

systemtest/driver/flow.py
```diff
@@ -80,27 +82,60 @@ def check_max_agents(flow_dict, max_agents_dict):
...
+    assert starting_times, "Could not read starting times from events file"
+    return starting_times
+
+
+def check_flow(
+    data_dict: dict,
+    max_agents_dict: dict,
+    starting_times_dict: dict,
+    outflow_dict: dict,
+):
```
Ozaq, line 104: Please add a doctoring giving a high level explanation what this does, including the mitigation for many agents passing over the measurement line in a single frame.

Ozaq (changes requested):

systemtest/driver/flow.py
zeroset: done.

systemtest/driver/flow.py
zeroset: done

systemtest/test_jpscore.py
zeroset: done

systemtest/driver/utils.py
```diff
@@ -91,3 +92,9 @@ def copy_files(*, sources: List[pathlib.Path], dest: pathlib.Path):
     for path in sources:
         if path.is_file():
             shutil.copy2(path, dest)
+
+def pairwise(iterable):
+    "s -> (s0,s1), (s1,s2), (s2, s3), ..."
```
Ozaq, line 97: This should be:
```
"""
s -> (s0,s1), (s1,s2), (s2, s3), ...
"""
```
While this is technically a doc string, i.e. a string literal as the first code inside the function, PEP-257 strongly recommends to use triple quoted strings for doc strings

systemtest/driver/inifile.py
```diff
@@ -37,3 +38,27 @@ def instanciate_tempalte(*, src: pathlib.Path, args: Dict, dest: pathlib.Path):
     tpl = jinja2.Template(src.read_text())
     content = tpl.render(args)
     dest.write_text(content)
+
+
+class TrafficConstraint(NamedTuple):
+    door_id: int
+    outflow: int
+    max_agents: int
```
Ozaq, line 46: I have two remarks about this code:

You could declare this as a dataclass which has similar semantics but is not a `NamedTuple` in case of `instanceof` calls.

We can stop using '-1' as indicator for "I have no value" and instead use 'None' instead. Additionally we can clearly communicate this by using `typing.Optional(int)`

E.g.:
```
from dataclasses import dataclass
from typing import Optional

@dataclass
class TrafficConstraint:
    door_id: int
    outflow: Optional(int) # <- this tells the type hints that 'None' is also acceptable
    max_agents: Optional(int)
```

systemtest/driver/inifile.py
```diff
@@ -37,3 +38,27 @@ def instanciate_tempalte(*, src: pathlib.Path, args: Dict, dest: pathlib.Path):
...
+        for door in tc.iter("door"):
+            traffic_constraints[
+                int(door.attrib["trans_id"])
+            ] = TrafficConstraint(
+                int(door.attrib["trans_id"]),
+                int(door.get("outflow", -1)),
+                int(door.get("max_agents", -1)),
+            )
+
+    assert traffic_constraints, "Could not parse traffic constraints"
```
Ozaq, line 63: I strongly prefer if only tests or 'check' functions assert. Imo this should either just return 'None' or throw an exception.

systemtest/driver/events.py
```diff
@@ -0,0 +1,16 @@
...
+    tree = ET.parse(events_file)
+    root = tree.getroot()
+    starting_times = {}
+    for event in root.iter("event"):
+        if event.attrib["state"] == "open" or event.attrib["state"] == "reset":
+            starting_times.setdefault(int(event.attrib["id"]), []).append(
+                float(event.attrib["time"])
+            )
+
+    assert starting_times, "Could not read starting times from events file"
```
Ozaq, line 15: I strongly prefer if only tests and 'check' functions assert. Imo this should either throw or return 'None'.

Ozaq (changes requested):

systemtest/driver/utils.py
zeroset: done

systemtest/driver/inifile.py
zeroset: done

systemtest/driver/events.py
zeroset: done

systemtest/driver/geometry.py
```diff
@@ -25,6 +18,6 @@ def check_traj_path_cross_line(path: ndarray, crossing_segment: Segment):
         segment = Segment(point1, point2)
         intersections = segment.intersection(crossing_segment)
         if intersections:
-            return True
+            return {"frame": p1[1], "intersections": intersections}
 
     return False
```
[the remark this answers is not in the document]
zeroset: done

systemtest/driver/inifile.py
zeroset: done

Ozaq (approved):

PR #969, opened 2022-03-18
Updates all models to use new geo data structures

Ozaq:
This change includes
* use of new neighborhood search for pedestrians
* use of new geometry


Ozaq:
Know issue: Right now reference tests will not pass because we have slight adjustments in the models so we have slight changes in the output. If the PR is otherwise ok I will update the reference results.

PR #1046, opened 2022-09-28
Make arguments for python api key only

chraibi:


python_bindings_jupedsim/bindings_jupedsim.cpp
```diff
@@ -155,8 +156,9 @@ PYBIND11_MODULE(py_jupedsim, m)
                double radius) {
                 JPS_VelocityModelBuilder_AddParameterProfile(w.handle, id, t, tau, v0, radius);
             },
+            py::kw_only(),
             py::arg("id"),
-            py::arg("t"),
+            py::arg("t_gap"),
```
Ozaq, line 161: please name it `time_gap` as you originally suggested

python_bindings_jupedsim/bindings_jupedsim.cpp
```diff
@@ -338,6 +342,7 @@ PYBIND11_MODULE(py_jupedsim, m)
             py::arg("model"),
             py::arg("geometry"),
             py::arg("areas"),
+            py::kw_only(),
```
Ozaq, line 345: I would like all parameters on simulation to require named arguments

Ozaq (changes requested):
Please also mandate named parameters on:
* `add_area`
* `switch_agent_profile`

python_bindings_jupedsim/bindings_jupedsim.cpp
chraibi: Addressed both comments in one commit

Ozaq (approved):

PR #1053, opened 2022-11-17
Agent Distribution with Bridson´s Poisson Disk Sampling

ChristianHirt:
Implements two Methods for Agent Distribution:
distribute till full & distribute by percentage
with Bridson´s Poisson Disk Sampling

todo:

- [x] adapt Tests for new Methods

### DIFF
--- a/python_modules/jupedsim/jupedsim/distributions.py
+++ b/python_modules/jupedsim/jupedsim/distributions.py
@@ -103,7 +103,8 @@ def distribute_by_number(*, polygon, number_of_agents, distance_to_agents, dista
     return grid.get_samples()
 
 
-def distribute_by_density(*, polygon, density, distance_to_agents, distance_to_polygon, seed=None, max_iterations=10000):
+def distribute_by_density(*, polygon, density, distance_to_agents, distance_to_polygon, seed=None,
+                          max_iterations=10000):
     """returns points randomly placed inside the polygon with the given density
 
         :param polygon: shapely polygon in which the agents will be placed
@@ -149,7 +150,8 @@ def __catch_wrong_inputs(polygon, center_point, circle_segment_radii, fill_param
                 j = j + 1
                 continue
             else:
-                raise OverlappingCirclesError(f"the Circle from {c_s_radius[0]} to {c_s_radius[1]} overlaps with others")
+                raise OverlappingCirclesError(
+                    f"the Circle from {c_s_radius[0]} to {c_s_radius[1]} overlaps with others")
 
 
 def distribute_in_circles_by_number(*, polygon, distance_to_agents, distance_to_polygon,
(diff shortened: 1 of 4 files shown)


python_modules/jupedsim/jupedsim/distributions.py
```diff
@@ -272,37 +274,63 @@ def distribute_in_circles_by_density(*, polygon, distance_to_agents, distance_to
...
-
-            iterations = 0
-        else:
-            iterations += 1
+        iteration = iteration + 1
+    if iteration >= max_iterations:
+        raise IncorrectParameterError("The first point could not be placed inside the polygon."
+                                      " Check if there is enough space for agents provided inside the polygon")
+
+    # while points are active a random reference point is selected
```
JetteSchumann, line 307: Reference of the algorithm is missing

python_modules/jupedsim/jupedsim/distributions.py
```diff
@@ -313,11 +341,20 @@ def distribute_by_percentage(*, polygon, percent, distance_to_agents, distance_t
...
     sample_amount = len(samples)
-    needed_amount = round(sample_amount * (percent/100))
-    return samples[:needed_amount]
+    needed_amount = round(sample_amount * (percent / 100))
+    np.random.seed(seed)
+    selected_samples = []
+    for i in range(needed_amount):
+        temp_sample = samples[np.random.randint(0, len(samples))]
+        selected_samples.append(temp_sample)
+        samples.remove(temp_sample)
```
JetteSchumann, line 354: Why can't you just do it the way before: `samples[:needed_amount]`?

python_modules/jupedsim/tests/test_distributions.py
```diff
@@ -235,11 +235,11 @@ def test_distribution_till_full_creates_correct_points():
     distance_to_agents = 2.0
     distance_to_polygon = 1.0
     set_seed = 1337
-    samples = distributions.distribute_til_full(polygon=polygon, distance_to_agents=distance_to_agents,
+    samples = distributions.distribute_till_full(polygon=polygon, distance_to_agents=distance_to_agents,
                                                 distance_to_polygon=distance_to_polygon,
                                                 seed=set_seed, max_iterations=750)
     # as many points created as intended
-    assert len(samples) == 18
+    assert len(samples) == 16
```
JetteSchumann, line 242: You could parametrize this test and try it with a higher number since 16 seems to be very small

python_modules/jupedsim/jupedsim/distributions.py
ChristianHirt: the next points always depend on the previously placed points.
So taking only the first points would create a cluster around the first placed point.

python_modules/jupedsim/tests/test_distributions.py
ChristianHirt: in this test only a few points are created since it is being tested if every placed point is correct.
This means for every point is determined if it is inside the polygon as well as the distance to every other point.
Having too many points will extend to runtime of the test exponentially.
However 16 is indeed not that high.
With that in Mind. How high should that number be / how long is a test allowed to last.

python_modules/jupedsim/jupedsim/distributions.py
```diff
@@ -272,37 +274,63 @@ def distribute_in_circles_by_density(*, polygon, distance_to_agents, distance_to
                                            numbers_of_agents=number_of_agents, seed=seed, max_iterations=max_iterations)
 
 
-def distribute_til_full(*, polygon, distance_to_agents, distance_to_polygon, seed=None, max_iterations=10000):
+def distribute_till_full(*, polygon, distance_to_agents, distance_to_polygon, seed=None, max_iterations=10_000, k=30):
     """returns as many randomly placed points as fit into the polygon
```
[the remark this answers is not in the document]
ChristianHirt: Algorithm noted

python_modules/jupedsim/jupedsim/distributions.py
ChristianHirt: reference to this paper added

python_modules/jupedsim/jupedsim/distributions.py
```diff
@@ -272,39 +274,68 @@ def distribute_in_circles_by_density(*, polygon, distance_to_agents, distance_to
...
-        else:
-            iterations += 1
+        iteration = iteration + 1
+    if iteration >= max_iterations:
+        raise IncorrectParameterError("The first point could not be placed inside the polygon."
+                                      " Check if there is enough space for agents provided inside the polygon")
+
+    # while points are active a random reference point is selected
+    while active:
+        ref_point = active[np.random.randint(0, len(active))]
```
JetteSchumann, line 311: Shouldn't it be: `np.random.randint(0, len(active)-1)` ?

python_modules/jupedsim/jupedsim/distributions.py
JetteSchumann: It might be better to shuffle the list and then just return `samples[:needed_amount]`

python_modules/jupedsim/tests/test_distributions.py
JetteSchumann: There is no correct number for tests 😄 

To keep the runtime low you could write several small tests, e.g. you could parametrize this test for different types of polygons and/or distances. So the number of created points must not be that high...

JetteSchumann (changes requested):

python_modules/jupedsim/jupedsim/distributions.py
ChristianHirt: no. 
random.randint(low, high=None, ... )
Return random integers from low (inclusive) to high (exclusive).

python_modules/jupedsim/tests/test_distributions.py
ChristianHirt: There are now three parametrized Tests testing if points are created correctly.
All of them are not taking to long. Tests should cover what needs to be tested.

python_modules/jupedsim/jupedsim/distributions.py
ChristianHirt: done

PR #1055, opened 2023-01-26
transfer demo file 1-3

ChristianHirt:
transfer old demo files from xml format into Python


ChristianHirt:
current demo file does not run due to RuntimeError: Point outside accessible area.

ChristianHirt:
this new version runs but never stops. the simulated agent keeps going back and forth.

PR #1087, opened 2023-03-31
Add Skript to convert AKNZ geometry to WKT

### LINKED ISSUE #1071: Convert AKNZ geometry to WKT
The attached old geometries need to be converted into WKT Geometry Collections (See: https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry), i.e. a collection of Polygons with holes.

There are two geometries attached, `Arrival` and `Evac`. `Evac` contains the geometry of `Arrival` plus a set of additional pathways for evacuation.

Geometry from `Arrival` shall be converted into a Geometry Collection.
Geometry from `Evac` shall be converted into a Geometry Collection that only contains the additional paths not present in `Arrival`

The resulting WKT strings are to be stored in
`examples/geometry/aknz_arrival.wkt` and `examples/geometry/aknz_evac.wkt` 

NOTE: The old JPS implementation allowed to have zero width walls, this is no longer supported. Any such wall has to be manually converted to a 5cm thick wall.

SUGGESTION: Use A semi automated process in python to read the old geometry and use shaply to write the requested WKT. See: https://shapely.readthedocs.io/en/stable/reference/shapely.to_wkt.html 

[Archive.zip](https://github.com/JuPedSim/jpscore/files/10837094/Archive.zip)

ChristianHirt:
Fixes #1071


ChristianHirt:
aknz_difference
This is the difference between the arrival and evacuation geometry.
you can see that it differs mainly in two new large emergency exits. But also a path on the left side is now a bit longer. Also inconspicuous but different are some coordinates above the path on the left side. these are very difficult to see here, but here some coordinates differ minimally.

scripts/read_geometry.py
```diff
@@ -0,0 +1,165 @@
+import xml.etree.ElementTree as ET
+from shapely import to_wkt, GeometryCollection
+from shapely.geometry import Polygon
+from shapely.ops import polygonize
+from shapely.ops import unary_union
+import matplotlib.pyplot as plt
+import geopandas as gpd
+
+
```
Ozaq, line 9: Please add some high level documentation about what this script is used for.

Ozaq (changes requested):
@ChristianHirt Please add some doc to the script, really just some high level description what this does and what it is used for. Just a couple of sentences will do. Otherwise this is good to go.

Ozaq (approved):

PR #1103, opened 2023-05-05
xml-parsing documentation

ChristianHirt:
adds some further documentation to the script to parse geometries
also removes main


Ozaq (approved):
Why did you remove main? How am I supposed to call the script?

PR #1125, opened 2023-06-20
fixing warnings

behrisch:


Ozaq (changes requested):
This breaks `-Wswitch` (part of all) We would no longer get a warning for an unhandled enum.

I suggest to use `__builtin_unreachable` instead. We use it already at a different place but it would be necessary to extract this first, see jupedsim.cpp:867 ff

Ozaq:
We can get this addressed on Monday if you are busy otherwise.

behrisch:
Is it  a problem of my compiler then? Because if the switch handles all cases then the compiler should detect it, shouldn't it?

Ozaq:
It looks like a short coming in GCC, I do not see warnings on apple-clang or clang.

`__builtin(unreachable)` or throwing a `std::runtime_error("unreachable")` pacifies GCC, I prefer the former.

Ozaq:
Closed in favour of PR #1128

PR #1140, opened 2023-07-13
Improve performance of linesegment point distance calculation

schroedtert:
Instead of using an own implementation of the line segment-point distance calculation, now use CGAL to compute the distance.


schroedtert:
Validation for using `CGAL::squared_distance` over own implementation:

image

Ozaq (approved):

PR #1154, opened 2023-07-27
Build python package with setup py

schroedtert:
To test the package you can run:

```bash
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ jupedsim==0.10.0a7
```

Link to test-pypi for preview: https://test.pypi.org/project/jupedsim/


schroedtert:
For discussion:
- Do we want to upload each push to the master branch somewhere or only releases?
- Should visdbg also be included in the package? It is now, but maybe would be better if it is optional

PR #1155, opened 2023-07-27
Update readme

Ozaq:


schroedtert (dismissed):

README.md
```diff
@@ -5,24 +5,17 @@
```
schroedtert, line 5: Shouldn't we also get rid of these links? As only `jpscore` still exists, but the documentation is heavily updated.

PR #1166, opened 2023-08-03
Add tool to update copyright headers

Ozaq:


scripts/Merge_Trajectories.py
```diff
@@ -1,5 +1,7 @@
 #! /usr/bin/env python3
 import argparse
+# Copyright © 2012-2023 Forschungszentrum Jülich GmbH
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
schroedtert, line 4: This looks weird shouldn't the copy right header be above the `import`?

scripts/copyrighter.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+    parser.add_argument("root", type=Path, default=Path.cwd())
+    return parser.parse_args()
+
+
+def main():
+    args = parse_arguments()
+    type_handlers = [
+        Handler(args.year, [".py"], "# {}", True),
+        Handler(args.year, [".hpp", ".cpp"], "// {}", False),
```
schroedtert, line 143: Before cpp copyright header started with `///`. Is there a reason for the change?

performancetest/performancetest/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright © 2012-2023 Forschungszentrum Jülich GmbH
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
```
schroedtert, line 3: Do empty files need a copyright header?

performancetest/large_street_network.py
```diff
@@ -1,5 +1,7 @@
 #! /usr/bin/env python3
 # Copyright © 2012-2022 Forschungszentrum Jülich GmbH
+# Copyright © 2012-2023 Forschungszentrum Jülich GmbH
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
schroedtert, line 4: Duplicated copy right header

performancetest/grosser_stern.py
```diff
@@ -1,5 +1,7 @@
 #! /usr/bin/env python3
 # Copyright © 2012-2022 Forschungszentrum Jülich GmbH
+# Copyright © 2012-2023 Forschungszentrum Jülich GmbH
+# SPDX-License-Identifier: LGPL-3.0-or-later
```
schroedtert, line 4: Duplicated copy right header

scripts/copyrighter.py
```diff
@@ -0,0 +1,156 @@
...
+            new_content += file_contents[1:]
+        return new_content
+
+    def _has_copyright_header(
+        self, file_contents: list[str], has_shebang: bool
+    ) -> bool:
+        if has_shebang:
+            return (
+                len(file_contents) >= 4
+                and self.header_pattern.search(file_contents[2]) != None
```
schroedtert, line 88: When only checking for `file_contents[2]` the following case is not handled properly:
```
#!/usr/bin/env python
# Copyright header
```
See annotated files. 
Won't it be better to check the first non empty line, if it is a copy right header?

PR #1168, opened 2023-08-04
Check if CMake and C++ compiler are available when building jupedsim

### LINKED ISSUE #1157: Extend setup.py with better error messages
Add check to find compiler and cmake and print some **readable** error message.

Idea:
1. Check for cmake
2. shell out a main.cpp and CMakeLists.txt
3. Check if this can be compiled.

schroedtert:
When installing JuPedSim via PyPI or locally, it will be checked whether CMake and a C++ Compiler are available, if not some error messages will be displayed to the users.

Closes #1157

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,8 @@ import re
 import shutil
 import subprocess
 import sys
+import tempfile
+import textwrap
 from pathlib import Path
 
 from setuptools import Extension, setup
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -32,6 +34,52 @@
...
+        cpp_file = tmp_dir + "/main.cpp"
+        with open(cpp_file, "w") as cp_file:
+            cp_file.write(simple_main)
+
+        simple_cmake = textwrap.dedent(
+            """
+            cmake_minimum_required(VERSION 3.19)
+            project(SimpleTest)
+
+            add_executable(simple_test main.cpp)
```
Ozaq, line 63: This needs to check for C++20

setup.py
```diff
@@ -32,6 +34,52 @@
 }
 
 
+def check_cmake():
+    try:
+        subprocess.run(["cmake", "--version"], check=True)
+    except:
+        return False
+    return True
```
Ozaq, line 42: This should capture the output and check for our minimum required cmake version

setup.py
```diff
@@ -43,6 +91,18 @@ def __init__(self, name: str, sourcedir: str = "") -> None:
 
 class CMakeBuild(build_ext):
     def build_extension(self, ext: CMakeExtension) -> None:
+        if not check_cmake():
+            raise ModuleNotFoundError(
+                "No CMake installation found on the system, "
```
Ozaq, line 96: Suggestion: "No CMake or no CMake >= 3.19 installation found on the system"

setup.py
```diff
@@ -43,6 +91,18 @@ def __init__(self, name: str, sourcedir: str = "") -> None:
...
+                "please install CMake to install JuPedSim."
+            )
+
+        if not check_cpp_compiler():
+            raise ModuleNotFoundError(
+                "Could not compile a simple C++ program, "
+                "please install a C++ compiler to install JuPedSim."
```
Ozaq, line 103: "please install a compiler with C++20 support"

setup.py
schroedtert: Added the check

setup.py
schroedtert: Now also check the cmake version

setup.py
schroedtert: Changed

setup.py
schroedtert: Changed

Ozaq (approved):

PR #1194, opened 2023-09-20
Api improvements

Ozaq:


python_modules/jupedsim/jupedsim/native/models.py
```diff
@@ -1,49 +1,32 @@
...
+@dataclass(kw_only=True)
+class GeneralizedCentrifugalForceModelParameters:
+    nu_ped: float
+    nu_wall: float
+    dist_eff_ped: float
+    dist_eff_wall: float
+    intp_width_ped: float
+    intp_width_wall: float
+    maxf_ped: float
+    maxf_wall: float
```
schroedtert, line 29: Why no default parameters here?

python_modules/jupedsim/jupedsim/native/models.py
```diff
@@ -1,49 +1,32 @@
...
 
 
 class GCFMModelAgentParameters:
```
schroedtert, line 32: Every other occurrence of GCFM has been renamed to full name, except this part. We should be consistent in the naming here.

schroedtert (commented):
One larger issue that i got: The trajectory data now only starts at frame 1. Which gives an empty image in the visualizer, as it starts at frame 0 but the agents only appear in frame 1 there.

Ozaq:
> One larger issue that i got: The trajectory data now only starts at frame 1. Which gives an empty image in the visualizer, as it starts at frame 0 but the agents only appear in frame 1 there.

Fixed

PR #1214, opened 2023-10-06
Update animation

chraibi:
- Add orientation lines to the circles.
- Better colors
- Add colorbar


chraibi:
The visualisation cell is added to the other notebooks in my other PR

PR #1221, opened 2023-10-09
Lane formation

chraibi:
Lane formation according to https://doi.org/10.1103/PhysRevE.94.032304


Ozaq:
Merged in a different PR

PR #1290, opened 2023-11-16
Add direct steering for agents

schroedtert:
This change will allow direct steering of agents, they will move towards

- [x] check if `DirectSteering`-Stage is the only one in a Journey, else throw error
- [x] ~~check if Agent is in a Journey with `DirectSteering` when setting waypoint, else throw error~~ Currently not possible, added notes to documentation.
- [x] ~~check if new waypoint in walkable area, else throw error.~~ Error will be thrown by the Routing Engine
- [x] rename `GenericAgent::waypoint` to `target`?

### DIFF
--- a/python_modules/jupedsim/jupedsim/agent.py
+++ b/python_modules/jupedsim/jupedsim/agent.py
@@ -30,6 +30,7 @@ class Agent:
         sim.agents_in_polygon(polygon)
 
     .. note ::
+
         You need to be aware that currently there are no checks done when setting
         properties on an Agent instance. For example it is possible to set an Agent position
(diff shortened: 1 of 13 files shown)


python_modules/jupedsim/jupedsim/agent.py
```diff
@@ -67,6 +67,14 @@ def orientation(self) -> tuple[float, float]:
         """Orientation of the agent."""
         return self._obj.orientation
 
+    @property
```
schroedtert, line 70: Add docu note: 
Can only set waypoint/target when in direct steering stage

PR #1295, opened 2023-11-22
Optimal Steps Model

schroedtert:


Ozaq:
This is a very rough draft of how we could integrate the Optimal Steps Model(OSM) into JuPedSim based on:

* [Natural discretization of pedestrian movement in continuous space](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.86.046108)
* [Dynamic stride length adaptation according to utility and personal space](https://arxiv.org/pdf/1401.7838v2.pdf)
* [How update schemes influence
crowd simulations](https://iopscience.iop.org/article/10.1088/1742-5468/2014/07/P07002)

The primary challenge is that JuPedSim expects all models to be generational, i.e. the state(n+1) of all agents is computed from the state(n). The OSM however uses an event-based approach where agents move one after another. 

The goal of this prototype was to:
A) Get a better understanding of the OSM and the respective [Vadere](https://gitlab.lrz.de/vadere/vadere) code.

B) Get an understanding of where architectural changes are required to integrate OSM

Outcome:

A) We now know "where to look" in Vadere for most implementation questions not answered by reading the publications above. What remains unclear is what the authors consider the "best" variation of OSM as the publications and Vadere offer a wide selection of variants.

We think we understand the OSM well enough to "be dangerous"

B) The prototype shows that we can embed the event based updates in an outer iteration based computation, i.e. When the simulation is requested to compute in time steps of 100ms we will handle all events that occur in the duration between t and t+100ms.

The prototype however is not working properly w.r.t to our agent removal handling and journey handling as this is tied to the main loops iteration time and now instead should act after each individual agents update. We did stop at this point because we can already see that the integration will not be finished in a few more days and we are not sure if the added complexity of providing a more general support for event based models if a worthwhile tradeoff.

We will need further discussion if we want to go th full way and integrate OSM

Issue #1337, opened 2024-03-08
Generalised Centrifugal Force Model broken


Ozaq:
Right now the Generalised Centrifugal Force Model is unusable.

To issues need to be addressed to solve this issue properly:

**Testing needs to be improved**
The Collision Free Speed Model is relatively well tested automatically by its use in the python notebooks of the documentation. Either we need to vary the use of models in the examples/tutorials/documentation or we need to create a few more general test cases (hint: RIMEA)

**Fixing the model**
There is nothing fundamentally wrong with the model but we did break the code when porting it from JuPedSim 0.8.x over to the new routing / neighbourhood queries etc...

PR #1370, opened 2024-04-22
fix repulsive forces

chraibi:
There was a confusion between the parameters of the repulsive function used to interpolate the functions.


Ozaq:
Closes #1337

PR #1422, opened 2024-05-31
disabling search state update to get all paths

behrisch:
I disabled the search state update only in the ComputeAllConsiderePaths function to show the effect. For the final fix it needs to be transferred to the ComputeAllWaypoints as well of course.


Ozaq:
We will pursue another approach entirely.

PR #1438, opened 2025-01-16
GCFM miscalculations when not initialised properly

chraibi:
I found a bug in the calculation of the ellipse distances. 
Since orientation is a unit vector, it directly provides the cosine and sine of the rotation angle.
However, at first the orientation might be a zero vector.

In this PR, I will check all equations related to ellipses and update their documentation. 

Hopefully, by then we will have a correctly  performing  gcfm.

### DIFF
--- a/libsimulator/src/Ellipse.cpp
+++ b/libsimulator/src/Ellipse.cpp
@@ -3,18 +3,42 @@
 #include "Ellipse.hpp"
 #include "Macros.hpp"
 
-// ellipse  semi-axis in the direction of the velocity
+/**
+ * @brief Calculates the semi-major axis (EA) of an ellipse based on the given speed.
+ *
+ * The ellipse adapts dynamically depending on the agent's speed. This method computes
+ * the semi-axis length in the direction of movement, reflecting how the ellipse
+ * elongates as speed increases.
+ *
+ * @param speed The current speed of the object (must be non-negative).
+ * @return The computed semi-axis length in the velocity direction.
+ *
+ */
 double Ellipse::GetEA(double speed) const
 {
     return Amin + speed * Av;
 }
 
-// ellipse semi-axis in the orthogonal direction of the velocity
+/**
+ * @brief Calculates the semi-minor axis (EB) of an ellipse orthogonal to the direction of velocity.
+ *
+ * This method computes the ellipse's semi-axis length perpendicular to the object's movement,
+ * allowing the ellipse to contract or expand based on the provided scaling factor.
+ *
+ * @param scale A scaling factor typically in the range \f$ [0, 1] \f$.
+ *        - \f$ \text{scale} = 0 \f$ → returns \f$ B_{\text{max}} \f$.
+ *        - \f$ \text{scale} = 1 \f$ → returns \f$ B_{\text{min}} \f$.
+ * @return The computed semi-axis length orthogonal to the velocity direction.
+ *
+ * @note Values of `scale` outside the \f$ [0, 1] \f$ range may produce unexpected results.
+ * @warning No input validation is performed on the `scale` parameter.
+ */
 double Ellipse::GetEB(double scale) const
 {
     const double deltaB = Bmax - Bmin;
     return Bmax - deltaB * scale;
 }
+
 
 double Ellipse::EffectiveDistanceToEllipse(
     const Ellipse& E2,
@@ -27,19 +51,16 @@ double Ellipse::EffectiveDistanceToEllipse(
     const Point& orientation_first,
     const Point& orientation_second) const
 {
-    Point R1, R2;
-    Point E1inE2, // center of E1 in coordinate system of E2
-        E2inE1;
-    E2inE1 = center_second.TransformToEllipseCoordinates(
-        center_first, orientation_first.x, orientation_first.y);
-    E1inE2 = center_first.TransformToEllipseCoordinates(
-        center_second, orientation_second.x, orientation_second.y);
-    // distance between centers of E1 and E2
-    const auto dist = (center_first - center_second).Norm();
-    R1 = this->PointOnEllipse(E2inE1, scale_first, center_first, speed_first, orientation_first);
-    R2 = E2.PointOnEllipse(E1inE2, scale_second, center_second, speed_second, orientation_second);
-    // effective distance
-    return dist - (center_first - R1).Norm() - (center_second - R2).Norm();
+    // Transform centers into each other's coordinate systems using orientation as cos(φ) and sin(φ)
+    Point E2inE1 = center_second.TransformToEllipseCoordinates(center_first, orientation_first.x, orientation_first.y);
+    Point E1inE2 = center_first.TransformToEllipseCoordinates(center_second, orientation_second.x, orientation_second.y);
+
+    // Find the closest points on each ellipse
+    Point R1 = this->PointOnEllipse(E2inE1, scale_first, center_first, speed_first, orientation_first);
+    Point R2 = E2.PointOnEllipse(E1inE2, scale_second, center_second, speed_second, orientation_second);
+
+    // Compute the shortest distance between the closest points
+    return (R1 - R2).Norm();
 }
 
 // input: P is a point in the ellipse world.
--- a/libsimulator/src/Ellipse.cpp
+++ b/libsimulator/src/Ellipse.cpp
@@ -3,58 +3,51 @@
 #include "Ellipse.hpp"
 #include "Macros.hpp"
 
-/**
- * @brief Calculates the semi-major axis (EA) of an ellipse based on the given speed.
- *
- * The ellipse adapts dynamically depending on the agent's speed. This method computes
- * the semi-axis length in the direction of movement, reflecting how the ellipse
- * elongates as speed increases.
- *
- * @param speed The current speed of the object (must be non-negative).
- * @return The computed semi-axis length in the velocity direction.
- *
- */
+/// Calculates the semi-major axis (EA) of an ellipse based on the given speed.
+///
+/// The ellipse adapts dynamically depending on the agent's speed. This method computes
+/// the semi-axis length in the direction of movement, reflecting how the ellipse
+/// elongates as speed increases.
+///
+/// @param speed The current speed of the object (must be non-negative).
+/// @return The computed semi-axis length in the velocity direction.
 double Ellipse::GetEA(double speed) const
 {
     return Amin + speed * Av;
 }
 
-/**
- * @brief Calculates the semi-minor axis (EB) of an ellipse orthogonal to the direction of velocity.
- *
- * This method computes the ellipse's semi-axis length perpendicular to the object's movement,
- * allowing the ellipse to contract or expand based on the provided scaling factor.
- *
- * @param scale A scaling factor typically in the range \f$ [0, 1] \f$.
- *        - \f$ \text{scale} = 0 \f$ → returns \f$ B_{\text{max}} \f$.
- *        - \f$ \text{scale} = 1 \f$ → returns \f$ B_{\text{min}} \f$.
- * @return The computed semi-axis length orthogonal to the velocity direction.
- *
- * @note Values of `scale` outside the \f$ [0, 1] \f$ range may produce unexpected results.
- * @warning No input validation is performed on the `scale` parameter.
- */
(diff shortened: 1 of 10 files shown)


libsimulator/src/Point.cpp
```diff
@@ -26,102 +25,56 @@ std::tuple<double, Point> Point::NormAndNormalized() const
...
 }
 
-/*rotate a two-dimensional vector by an angle of theta
-
-Rotation-matrix=[cos(theta)  -sin(theta)]
-                [ sin(theta)  cos(theta)]
-
-*/
+/**
+ * Helper method to rotate a point around the origin counterclockwise
```
Ozaq, line 73:
```suggestion
 * CCW rotate Point (interpreted as vector) around origin.
```

Ozaq (changes requested):
Hey @chraibi if I understand it correctly the issue only manifests when adding agents, when their orientation vector is erroneously [0,0], right? 

If the above is the case, I think it would make more sense to require users to provide a unit length orientation vector as part of the construction? And fail on construction, i.e as early as possible.

Further, the change you made would still allow users to set an initial orientation of [0,0] but would abort(!) the program at runtime. Not what we generally want as this kills the whole process. Instead we should throw an exception and always do the check, this way we can properly inform the user.

Also: could you please add a regression test that exhibits the problem? Let me know if I should lend you a hand in that.

chraibi:
@Ozaq  
> If the above is the case, I think it would make more sense to require users to provide a unit length orientation vector as part of the construction? And fail on construction, i.e as early as possible.

Yes, that's what I just realized. However, GCFM is the only model that works with Ellipses. Since, we initialize the orientation to (0,0) and this means that cos(phi)=0 and sin(phi)=0 at the same time.

You are right. The problem can easily be solved by having the user initialize the orientation.
Or so I thought, before debugging:
I add an agent so
```
jps.GeneralizedCentrifugalForceModelAgentParameters(
            orientation=(1, 0),
            journey_id=journey_id,
            stage_id=exit_id,
            v0=1,
            position=pos,
            # a_min=0.1,
            # a_v=0,
            # b_min=0.1,
            # b_max=0.1,
        )
```
and this is what I get
Screenshot 2025-01-17 at 15 47 05


I prefer not to make these calls and rather validate the function input like so:

```
{
    if((std::abs(cphi) < 1e-6 && std::abs(sphi) < 1e-6) ||
       std::abs(cphi * cphi + sphi * sphi - 1.0) > 1e-6) {
        throw std::invalid_argument(
            "Invalid rotation inputs: vector is zero or not normalized (cphi^2 + sphi^2 != 1). "
            "Values: cphi=" +
            std::to_string(cphi) + ", sphi=" + std::to_string(sphi));
    }
```



> Further, the change you made would still allow users to set an initial orientation of [0,0] but would abort(!) the program at runtime. Not what we generally want as this kills the whole process. Instead we should throw an exception and always do the check, this way we can properly inform the user.

I did not try yet to solve the problem. 
In this Draft PR, I'm just trying, very cautiously, to go through the functions and update their documentations.
 
> Also: could you please add a regression test that exhibits the problem?
Yes, I updated one test by adding a second agent. This already produces the problem.

Ozaq:
>However, GCFM is the only model that works with Ellipses. Since, we initialize the orientation to (0,0) and this means that cos(phi)=0 and sin(phi)=0 at the same time.

True, but an orientation of (0,0) is also a constraint violation because we say "Orientation is a unit length vector", we could also have used an angle in the API when creating the agent and then convert it into a vector internally. This might be a useful API change for a potential 2.0. What do you think?

chraibi:
@Ozaq  I find orientation is more intuitive and easier to use than angle. For now, I check for it during the initialization phase.
Just need to figure out how to check for it again during a running simulation when the user decides to change it again. 

I think this PR solves already the problem I had before.

libsimulator/src/Simulation.cpp
```diff
@@ -8,6 +8,7 @@
 #include "OperationalModel.hpp"
 #include "Stage.hpp"
 #include "Visitor.hpp"
+#include <Logger.hpp>
```
Ozaq, line 11: Remove

libsimulator/src/Simulation.cpp
```diff
@@ -177,17 +178,33 @@ BaseStage::ID Simulation::AddStage(const StageDescription stageDescription)
 
 GenericAgent::ID Simulation::AddAgent(GenericAgent&& agent)
 {
+
```
Ozaq, line 181: Remove

libsimulator/src/Point.cpp
```diff
@@ -26,102 +34,51 @@ std::tuple<double, Point> Point::NormAndNormalized() const
...
-Rotation-matrix=[cos(theta)  -sin(theta)]
-                [ sin(theta)  cos(theta)]
-
-*/
+/**
+ * CCW rotate Point (interpreted as vector) around origin.
+ * @param ctheta Cosine of rotation angle
+ * @param stetha Sine of rotation angle
+ * @return Rotated point
+ */
```
Ozaq, line 81: Please follow the established documentation style and use `///` to indicate doc-style comments.

libsimulator/src/Point.cpp
```diff
@@ -26,102 +34,51 @@ std::tuple<double, Point> Point::NormAndNormalized() const
...
+ * 3. Add ellipse center coordinates to translate: + [xc ; yc]
+ *
+ * @param center The ellipse center coordinates (xc,yc) in Cartesian system S_1
+ * @param cphi   cos(phi), where phi is the angle between S_1 and S_2
+ * @param sphi   sin(phi)
+ * @return       Point in Cartesian coordinates (S_1)
+ */
 Point Point::TransformToCartesianCoordinates(const Point& center, double cphi, double sphi) const
 {
+
```
Ozaq, line 71: Remove

libsimulator/src/Point.cpp
```diff
@@ -26,102 +34,51 @@ std::tuple<double, Point> Point::NormAndNormalized() const
...
+ * The transformation follows these steps:
+ * 1. Take point P(x,y) in ellipse system S_2
+ * 2. Rotate by -phi to align with Cartesian axes: M(-phi) * [x_2 ; y_2]
+ * 3. Add ellipse center coordinates to translate: + [xc ; yc]
+ *
+ * @param center The ellipse center coordinates (xc,yc) in Cartesian system S_1
+ * @param cphi   cos(phi), where phi is the angle between S_1 and S_2
+ * @param sphi   sin(phi)
+ * @return       Point in Cartesian coordinates (S_1)
+ */
```
Ozaq, line 68: Please follow the established documentation style and use `///` to indicate doc-style comments.

libsimulator/src/Point.cpp
```diff
@@ -26,102 +34,51 @@ std::tuple<double, Point> Point::NormAndNormalized() const
...
+ *
+ * Steps:
+ * 1. **Translation:** Move the point relative to the ellipse center.
+ * 2. **Rotation:** Rotate by angle -phi to align with S_2 axes.
+ *
+ * @param center  The center of the ellipse in Cartesian coordinates (xc, yc)
+ * @param cphi    Cosine of rotation angle phi
+ * @param sphi    Sine of rotation angle phi
+ * @return        Point transformed into the ellipse-local frame (S_2)
+ */
```
Ozaq, line 49: Please follow the established documentation style and use `///` to indicate doc-style comments.

libsimulator/src/Point.cpp
```diff
@@ -1,12 +1,20 @@
...
 
+bool Point::isInvalidOrientation() const
+{
+    if((std::abs(x) < 1e-6 && std::abs(y) < 1e-6) || (std::abs(NormSquare() - 1.0) > 1e-6)) {
+        LOG_WARNING("Orientation vector ({}, {}) is not normalized, correcting it.", x, y);
+        return true;
+    }
+    return false;
+}
+
```
Ozaq, line 17: I think this should be renamed to 'isZeroLength' because the current name implies that the Point has to be used as orientation for this function to make sense.

Also the warning message is misleading because the no correction is taking place and in general this is not something we should "warn" the user about. For the warning we are operating on at least one abstraction level below the code that should log a warning, if at all.

libsimulator/src/Point.cpp
```diff
@@ -1,12 +1,20 @@
 // Copyright © 2012-2024 Forschungszentrum Jülich GmbH
 // SPDX-License-Identifier: LGPL-3.0-or-later
 #include "Point.hpp"
-
-#include "Macros.hpp"
+#include <cassert>
```
Ozaq, line 4: This is unused.

libsimulator/src/Ellipse.cpp
```diff
@@ -27,25 +66,35 @@ double Ellipse::EffectiveDistanceToEllipse(
...
+ * \( P \).
+ *
+ * **Behavior:**
+ * - If \( P \) is very close to the ellipse center, it defaults to returning the point \((a, 0)\)
+ * on the ellipse.
+ * - Otherwise, it scales the direction from the center to \( P \) by the ellipse’s semi-major and
+ * semi-minor axes.
+ *
+ * @return Point on the ellipse boundary, transformed into the global coordinate system.
+ */
```
Ozaq, line 97: Please follow the established documentation style and use /// to indicate doc-style comments. Please omit the brief because we usually use Doxygen with auto-brief, i.e. First line / first sentence is the brief.

libsimulator/src/Ellipse.cpp
```diff
@@ -3,19 +3,58 @@
...
+ *    Transforms the center of each ellipse into the local coordinate system of the other
+ *    using their orientation vectors.
+ *
+ * 2. **Closest Point Determination:**
+ *    Identifies the closest point on each ellipse to the other ellipse in their respective
+ *    coordinate systems.
+ *
+ * 3. **Effective Distance Calculation:**
+ *    Calculates the Euclidean distance between these two closest points on the ellipses.
+ */
```
Ozaq, line 57: Please follow the established documentation style and use /// to indicate doc-style comments.

libsimulator/src/Ellipse.cpp
```diff
@@ -3,19 +3,58 @@
...
+ * allowing the ellipse to contract or expand based on the provided scaling factor.
+ *
+ * @param scale A scaling factor typically in the range \f$ [0, 1] \f$.
+ *        - \f$ \text{scale} = 0 \f$ → returns \f$ B_{\text{max}} \f$.
+ *        - \f$ \text{scale} = 1 \f$ → returns \f$ B_{\text{min}} \f$.
+ * @return The computed semi-axis length orthogonal to the velocity direction.
+ *
+ * @note Values of `scale` outside the \f$ [0, 1] \f$ range may produce unexpected results.
+ * @warning No input validation is performed on the `scale` parameter.
+ */
```
Ozaq, line 35: Please follow the established documentation style and use /// to indicate doc-style comments.

libsimulator/src/Ellipse.cpp
```diff
@@ -3,19 +3,58 @@
...
+ * @brief Calculates the semi-major axis (EA) of an ellipse based on the given speed.
+ *
+ * The ellipse adapts dynamically depending on the agent's speed. This method computes
+ * the semi-axis length in the direction of movement, reflecting how the ellipse
+ * elongates as speed increases.
+ *
+ * @param speed The current speed of the object (must be non-negative).
+ * @return The computed semi-axis length in the velocity direction.
+ *
+ */
```
Ozaq, line 16: Please follow the established documentation style and use /// to indicate doc-style comments.

systemtest/test_model_properties.py
```diff
@@ -135,7 +135,14 @@ def test_set_model_parameters_generalized_centrifugal_force_model(
...
         position=(1, 1),
     )
+    agent2 = jps.GeneralizedCentrifugalForceModelAgentParameters(
+        journey_id=journey_id,
+        stage_id=wp,
+        position=(3, 1),
+    )
+
     agent_id = sim.add_agent(agent)
+    sim.add_agent(agent2)
```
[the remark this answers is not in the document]
chraibi: By placing two pedestrians the test fails. 
This reproduces the problem with 0 orientation.

Ozaq (approved):

PR #1489, opened 2025-09-20
Add path diversity to reduce agent clustering in narrow corridors

chraibi:
The shortest path algorithm causes all agents to follow identical routes, creating unrealistic single-file movement through corridors and corners. This change introduces agent-specific bias in the funnel algorithm's clearance calculations, generating natural path variation while maintaining safe navigation.

Changes:
- Modified straightenPath() to accept path bias parameter (-0.8 to 0.8)
- Added deterministic bias generation based on agent ID hash
- Agents now prefer different sides of available corridor space
- Maintains minimum safety clearance while utilizing extra space for variation
- Cache calculated path. Recalculate if agent is near its waypoint.


chraibi:
trajectoriesuturn_new_n_22 sqlite

chraibi:
Caching is a bit faster (for N=22)
image

chraibi:
plots

libsimulator/src/RoutingEngine.cpp
```diff
@@ -19,16 +19,131 @@
...
+
+        for(int i = 0; i < 3; ++i) {
+            const auto& p = fit->vertex(i)->point();
+            out << CGAL::to_double(p.x()) << "," << CGAL::to_double(p.y()) << " ";
+        }
+        out << "\n";
+    }
+
+    std::cout << "Exported triangles to " << filename << std::endl;
+}
```
Ozaq: Check the visualiser code how to do this in python, unless im mistaken there is no need to do this from C++.

Ozaq (changes requested):

PR #1490, opened 2025-10-11
Fix GH Actions

Ozaq:


schroedtert (approved):
Looks good to me. Just raised a different question for me: do we want to have a nightly/weekly build? Would make it easier to detect pipeline issues

PR #1500, opened 2025-10-27
batch serializer - Updating sqlite_serialization.py to support buffered trajectory writing

FabianPlum:
At the moment, the serialisation (via sqilte)is a major performance bottleneck as we write out the trajectory data at every iteration (essentially every time `simulation_cfsm.iterate()` is called, unless steps are specified otherwise).

This PR adds buffered output writing, which can be enabled when configuring the `trajectory_writer`

e.g.

```
simulation_cfsm = jps.Simulation(
    model=jps.CollisionFreeSpeedModel(),
    geometry=geometry,
    trajectory_writer=jps.SqliteTrajectoryWriter(
        output_file=pathlib.Path(trajectory_file),
        buffer_in_memory=True
    ),
)
```

When buffering is enabled, by default the buffer is flushed to the disk every `max_buffered_frames: int = 1000` iterations. This optional argument can be supplied alongside `buffer_in_memory` above.

This implies, that after the simulation has ran through we need to call 'close()' (or alternatively 'flush()' if the file should remain opened) to ensure the buffer is cleared and the complete trajectories are written to the specified output file.

```
while (
    simulation_cfsm.agent_count() > 0
    and simulation_cfsm.iteration_count() < 20000
):
    simulation_cfsm.iterate()

simulation_cfsm._writer.close()
```

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1500/

### DIFF
--- a/python_modules/jupedsim/jupedsim/sqlite_serialization.py
+++ b/python_modules/jupedsim/jupedsim/sqlite_serialization.py
@@ -200,7 +200,7 @@ class SqliteTrajectoryWriter(TrajectoryWriter):
                 self._con.close()
             finally:
                 self._con = None  # type: ignore[assignment]
-    
+
     def every_nth_frame(self) -> int:
         return self._every_nth_frame
 


python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -35,6 +35,8 @@ def __init__(
         *,
         output_file: Path,
         every_nth_frame: int = 4,
+        buffer_in_memory: bool = False,
+        max_buffered_frames: int = 1000,
```
Ozaq: With no manual buffering this probably should one additional parameter: `commit_every_nth_write: int = 100` Or some other sensible default level.

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -44,19 +46,28 @@ def __init__(
...
             every_nth_frame: int
                 indicates interval between writes, 1 means every frame, 5 every 5th
-
-        Returns:
-            SqliteTrajectoryWriter
+            buffer_in_memory: bool
+                if True, iteration data is kept in RAM and written only when the
+                buffer reaches max_buffered_frames or when flush() / close() is called.
+            max_buffered_frames: int
+                maximum number of frames to keep in RAM before auto-flush.
```
Ozaq: Update accordingly to previous comment.

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -114,12 +125,17 @@ def begin_writing(self, simulation: Simulation) -> None:
         except sqlite3.Error as e:
             cur.execute("ROLLBACK")
```
Ozaq: We can no longer do this if journal mode is disabled, we can just remove this line.

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -174,12 +191,49 @@ def write_iteration_state(self, simulation: Simulation) -> None:
...
+        """Flush any buffered frames to disk in a single transaction.
+
+        Safe to call multiple times; if the buffer is empty, this won't execute.
+        """
+        if not self._buffer_in_memory:
+            return
+        if self._buffered_frame_count == 0:
+            return
+        cur = self._con.cursor()
+        try:
```
Ozaq: I don't think we need this, since with no hand written buffering this should(TM) be only a `execute("COMMIT")`

Ozaq (changes requested):
Like the changes. We trade off safety with performance in a place where we do not have any strong safety requirement.

I also would not bother to keep the current behaviour, we are working on a 2.0 release anyways so this is an acceptable behaviour change.

Until then @chraibi you can cherry-pick this class into your experiments on the FJZ OpenStack. Let me know if you need some help with that.

FabianPlum:
@chraibi @Ozaq Do we wanna kick the `every_nth_frame: int = 4` altogether?
I don't know for which cases this was added (perhaps when you have long running sims and don't need the fine temporal res?) but if it's just to save IO operations, we could kick it?

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -114,12 +125,17 @@ def begin_writing(self, simulation: Simulation) -> None:
         except sqlite3.Error as e:
             cur.execute("ROLLBACK")
             raise TrajectoryWriter.Exception(f"Error creating database: {e}")
+        finally:
+            if self._buffer_in_memory:
+                # start a transaction for subsequent writes
+                cur.execute("BEGIN")
```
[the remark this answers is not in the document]
FabianPlum: Do we actually need to call BEGIN again, after the initial call to `begin_writing`?
I just removed the call completely, with only COMMIT in `flush()`, and things appear to execute fine.
Not sure if we are dropping data somewhere, but all looks good at first glance.

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -174,12 +191,49 @@ def write_iteration_state(self, simulation: Simulation) -> None:
...
+            if not self._buffer_in_memory:
+                cur.execute("COMMIT")
+            else:
+                # Trigger flush if buffer full
+                self._buffered_frame_count += 1
+                if self._buffered_frame_count >= self._max_buffered_frames:
+                    self.flush()
+                return
+        except sqlite3.Error as e:
+            cur.execute("ROLLBACK")
```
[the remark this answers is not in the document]
FabianPlum: removed them all

Ozaq:
`every_nth_frame` is orthogonal to the `commit_every_nth_write` as in the first essentially says only store information about every nth iteration, simply because there is not so much happening in 10ms (dT for most models), larger scenarios are still good with recording the output at dT=1s (simulation is always done at dT=10ms, unless configured differently)

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -44,19 +45,24 @@ def __init__(
...
             raise TrajectoryWriter.Exception("'every_nth_frame' has to be > 0")
         self._every_nth_frame = every_nth_frame
-        self._con = sqlite3.connect(self._output_file, isolation_level=None)
+        # sqlite connection (with disabled autocommit mode); we still use explicit BEGIN/COMMIT
+        self._con = sqlite3.connect(self._output_file)#, isolation_level=None)
         # Don't wait for the OS to persist data
         self._con.execute("PRAGMA synchronous=OFF;")
         # Don't allow rollbacks (we don't have need for it)
         self._con.execute("PRAGMA journal_mode=OFF;")
+        #self._cur = self._con.cursor()
```
Ozaq: Please remove the commented out code

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -44,19 +45,24 @@ def __init__(
...
+            commit_every_nth_write: int
+                number of frames to keep in RAM before writing them to disk, triggered vai flush().
         """
         self._output_file = output_file
         if every_nth_frame < 1:
             raise TrajectoryWriter.Exception("'every_nth_frame' has to be > 0")
         self._every_nth_frame = every_nth_frame
-        self._con = sqlite3.connect(self._output_file, isolation_level=None)
+        # sqlite connection (with disabled autocommit mode); we still use explicit BEGIN/COMMIT
+        self._con = sqlite3.connect(self._output_file)#, isolation_level=None)
```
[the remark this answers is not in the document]
FabianPlum: done

python_modules/jupedsim/jupedsim/sqlite_serialization.py
FabianPlum: Done

FabianPlum:
@chraibi @Ozaq Just so I understand correctly, `docs/developer_guide/custom_writers.rst` (or developer_guide in general) is not something we have yet but rather a documentation resource we want to add?

chraibi:
> @chraibi @Ozaq Just so I understand correctly, `docs/developer_guide/custom_writers.rst` (or developer_guide in general) is not something we have yet but rather a documentation resource we want to add?

Yes, we don't have this yet.

python_modules/jupedsim/jupedsim/sqlite_serialization.py
```diff
@@ -44,20 +45,23 @@ def __init__(
...
         self._every_nth_frame = every_nth_frame
-        self._con = sqlite3.connect(self._output_file, isolation_level=None)
+        self._con = sqlite3.connect(self._output_file)
         # Don't wait for the OS to persist data
         self._con.execute("PRAGMA synchronous=OFF;")
         # Don't allow rollbacks (we don't have need for it)
         self._con.execute("PRAGMA journal_mode=OFF;")
 
+        # Buffering options
+        self._commit_every_nth_write = int(commit_every_nth_write) if commit_every_nth_write > 0 else 1
```
Ozaq: The general philosophy in JuPedSim is that we try not to anticipate what users want or try to fix user input errors. This needs to throw an exception if `commit_every_nth_write` is less that 1

FabianPlum:
Ran a few more tests and when we call '_writer.close()' with an empty buffer we get:

```
OperationalError                          Traceback (most recent call last)
Cell In[10], line 10
      7 while simulation.agent_count() > 0:
      8     simulation.iterate()
---> 10 simulation._writer.close()

File ~\.conda\envs\jupedsim-pip\Lib\site-packages\jupedsim\sqlite_serialization.py:196, in SqliteTrajectoryWriter.close(self)
    194 """Flush buffer and close DB connection. Call at simulation end."""
    195 cur = self._con.cursor()
--> 196 cur.execute("COMMIT")
    197 if self._con:
    198     try:

OperationalError: cannot commit - no transaction is active
```

So `cur.execute("COMMIT")` is not a no-op when there is nothing to commit here.
To handle this cleanly, I just added the same check as we use in the iteration step and simply don't flush the buffer when there is nothing to commit:

```
"""Flush buffer and close DB connection. Call at simulation end."""
if self._buffered_frame_count != 0:
    cur = self._con.cursor()
    cur.execute("COMMIT")
```

chraibi (approved):

FabianPlum:
so sorry for the formatting jazz.. Just reformatted with ruff rather than trying to manually address these issues...

Ozaq:
Will be merged from #1504

PR #1516, opened 2025-11-20
Fix freezing AVM-agents

chraibi:
AVM-Agents and for that matter also CFM-agents tend to freeze in certain situations. 
It does not happen often, but there are always some spacial configurations, where this might still happen. 

The reason for that is the speed function can return a zero speed. In that case, agents might not get out of their deadlock and "freeze" forever, which is really a sad scenario if you imagine it.
 
Screenshot 2025-12-13 at 13 53 02

In this PR I'm trying to:
- simplify the wall handling.
- and add a little randomness to the speed function to remove these freezings and still not mess-up with the theoretical definition of the model.

Fixes #1473 

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1516/

### DIFF
--- a/examples/avm-sharp-corners.py
+++ b/examples/avm-sharp-corners.py
@@ -128,6 +128,5 @@ if simulation.agent_count():
     print(
         f"\033[31mAgents remaining: {simulation.agent_count()}. (should be 0)\033[0m"
     )
-    break
 else:
     print("\033[32mAll agents have exited the simulation successfully.\033[0m")


libsimulator/src/AnticipationVelocityModel.cpp
```diff
@@ -311,67 +311,41 @@ Point AnticipationVelocityModel::HandleWallAvoidance(
...
+            const auto projectedDirection = modifiedDirection - normalTowardAgent * dotProduct;
+
+            if(distance <= criticalWallDistance) {
+                modifiedDirection = projectedDirection + normalTowardAgent * pushoutStrength;
+            } else {
+                // Soft influence: blend between original and projected based on distance
+                const double blendFactor =
+                    (influenceDistance - distance) / (influenceDistance - criticalWallDistance);
+                modifiedDirection =
+                    modifiedDirection * (1.0 - blendFactor) + projectedDirection * blendFactor;
```
FabianPlum: Should we clamp the blend factor to [0,1] ?

examples/avm-sharp-corners.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+for coords in default_geometry_coordinates:
+    poly = Polygon(coords)
+    if not poly.is_valid:
+        print("Invalid polygon:", coords)
+    if not poly.is_simple:
+        print("Non-simple polygon:", coords)
+
+print("Geometry is fine!")
```
FabianPlum: Regardless of whether the Geometry is *actually fine* (even if Invalid polygons or non-simple polygons are found) we always inform the user the Geometry is fine

examples/avm-sharp-corners.py
```diff
@@ -0,0 +1,133 @@
...
+
+while (
+    simulation.agent_count() > 0 and simulation.iteration_count() < niterations
+):
+    simulation.iterate()
+
+simulation._writer.close()
+
+print(
+    f"Simulation finished after {simulation.iteration_count()} iteration ({simulation.elapsed_time()} seconds.)"
```
FabianPlum: super minor -> this should be 'iteration**s**', plural, unless everyone leaves in the first iteration.

chraibi:
> So when you run the example, do you consistently get results with all agents having left the sim?

It was not the case before, but now I do.
image

FabianPlum:
> > So when you run the example, do you consistently get results with all agents having left the sim?
> 
> It was not the case before, but now I do. image

Sweet!

PR #1519, opened 2025-11-21
Update examples

chraibi:
closes #1518 

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1519/


FabianPlum (changes requested):
Please update the order of closing and print calls, just so that those are consistent. The rest looks great and I think we can merge this.

FabianPlum (approved):
All looks good here!

PR #1528, opened 2026-02-13
Fix issues

Ozaq:
Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1528/

### DIFF
--- a/python_modules/jupedsim/jupedsim/simulation.py
+++ b/python_modules/jupedsim/jupedsim/simulation.py
@@ -277,7 +277,7 @@ class Simulation:
         ):
             model = py_jps.GeneralizedCentrifugalForceModelState(
                 speed=parameters.speed,
-                desired_direction=parameters.orientation,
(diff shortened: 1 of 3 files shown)


schroedtert (commented):
First quick review, need to check the visitor change for the model formatting and pipeline changes in more detail.

python_modules/jupedsim/jupedsim/simulation.py
```diff
@@ -277,7 +277,7 @@ def add_agent(
         ):
             model = py_jps.GeneralizedCentrifugalForceModelState(
                 speed=parameters.speed,
-                desired_orientation=parameters.orientation,
+                desired_direction=parameters.orientation,
```
schroedtert: Shouldn't we then also adapt the parameter name too? Or would that break the interface?

Ozaq: we can break the interface its 2.0 prep still

PR #1537, opened 2026-03-06
Add CollisionFreeSpeedModelV3 with rotational steering

chraibi:
Preview: https://pedestriandynamics.github.io/jupedsim/pull-requests/1537/

- Add `CollisionFreeSpeedModelV3`, a minimal rotational steering variant derived from V2
- Direction is updated by rotating a reference direction away from the most influential
    forward neighbor, instead of summing repulsion vectors
- Speed uses spacing-based control with `agentBuffer` stand-off and a tiny reverse floor
    (`-0.01 m/s`) to release contact blockages
- Temporal heading relaxation (`TauTheta = 0.3s`) damps rapid left-right oscillations
- Anisotropic neighbor weight uses uniform linear-exponential decay on both axes;
    directionality comes from the `r_x / r_y` ratio alone
- Boundary influence bends the reference direction only (not summed into final velocity as in V2)

  ## Key differences from V2

  | Aspect | V2 | V3 |
  |---|---|---|
  | Neighbor avoidance | Vector-sum repulsion | Rotational steering (single dominant neighbor) |
  | Boundary role | Summed into velocity | Bends reference direction only |
  | Heading state | None | Persistent, relaxed toward target |
  | Speed floor | `0.0` | `-0.01 m/s` (deterministic reverse) |
  | Agent buffer | N/A | Shifts speed-zero to positive spacing |
  | Anisotropie | Distances are equally important | distances from the side are less important|
  | Speed function | depends on the distance in direction of movement | blend with direction in direction of goal|
  
Screenshot 2026-03-04 at 12 34 05

  
### Debugging Demo

[Demo](https://www.chraibi.de/CSM2/)
  

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1537/


Ozaq:
@chraibi IIRC you need this for a new publication? If yes we should mark this for a 1.4 release

chraibi:
> @chraibi IIRC you need this for a new publication? If yes we should mark this for a 1.4 release

Yes. I would also release 1.4 with both new models even.

PR #1545, opened 2026-03-14
Add WarpDriver operational model

chraibi:
## Summary

- Add WarpDriver operational model (Wolinski, Lin, Pettré 2016) across all three layers: C++ core, pybind11 bindings, Python wrapper
- Probabilistic collision field model: agents project trajectories, evaluate collision probability via warped intrinsic fields, and adjust velocity through gradient descent
- Random lateral perturbation on trajectory samples to break symmetry in head-on encounters (speed model, no repulsion forces)
- Boundary steering to keep agents away from walls

- [ ] Add model to model-notebook.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1545/


chraibi:
benchmark_models

chraibi:
Although the mode calculates fewer steps, it's overall a bit slower most other models.
Depending on the scenario, this might or might not hurt. 

In this PR the model will be introduced. Optimised, if needed, in another step. 

image

chraibi:
Warpdriver seems to be able to handle larger dt values. So although it is slower per time step (for the same dt) that fact that it need less iterations, and given that we can use larger dt values, overall it is faster..
(Depending on the scenario)

image

chraibi:
OP#10610

chraibi:
@ralfleibold should this PR be closed in favor of #1559 ?

ralfleibold:
> @ralfleibold should this PR be closed in favor of #1559 ?
I don't think so. The changes are too massive to easily merge the rls-1.4 branch at the end to master again. I think we best do in this PR again. The challenge will be to figure out what has changed on rls-1.4 (on the backport) and to apply the same changes here again.

ralfleibold (approved):
Thanks, Mohcine.

PR #1554, opened 2026-04-20
Fix issue 1552: Removed Agents does not work via Python API anymore.

ralfleibold:
Fix issue that `removed_agents()` does not work in Python API anymore (once there are removed agents). The issue is that UniqueID<T> is unknown on Pyhon level.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1554/


pitsteinbach (approved):
looks good to me

PR #1555, opened 2026-04-20
Add Timer and profiler traces

pitsteinbach:
I added/extended timer functionality to the previous Performance Statistics class.

The main interface is using strings to push and pop timings to the class and store them in a `std::map`.
Pushing starts the timer, whereas pop stops the timer and computes/sums up the duration spend with that certain key.

The legacy functions that time the main iteration loop as well as the Operational Model are kept and use the new infrastructure in the backend. Additionally, the `std::map` can be exported to Python, on the other hand Python can also push and pop timers in the C++ instance. The Python API was extended, if there are remaining applications let me know.

The timer is running now by default ad it probes the C++ functions within the main iteration loop as well as adding agents to the simulation, adding journeys or stages. The Python constructor and destructor of a `Simulation` object are used to mark the start and end of a simulation. The destructor of the Simulation object also prints the timings by default to std out.

In addition to a timer, I also added a project that allows to write Traces that can be read by the builtin profiler of Chrome, the tracing is off by default, but can be enabled via Python and writes the traces to a file currently called by default `jps_simualtor.json`.

The tracing information could be refined in the future.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1555/


chraibi:
OP#10563

ralfleibold:
At first glance I do not find License information on the "Profiler" project included. Are we sure we can just use this code?

pitsteinbach:
True, another option would be Minitrace https://github.com/hrydgard/minitrace it has a MIT License and the project states Windows compatibility which is something that Profiler has some issues with apparently..

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,67 @@
...
-    uint64_t iterate_duration{};
-    uint64_t op_dec_system_run_duration{};
-    bool enabled{false};
+    mutable uint64_t last_iteration_duration{0};
+    mutable uint64_t last_operational_duration{0};
+    bool enable_tracing{false};
+#ifdef BUILD_PROFILER    
+    std::string prof_filename{"jps_simulator.json"};
+    std::shared_ptr<Profiler> profiler{};
+    std::map<std::string, std::shared_ptr<DurationEvent>> event_map{};
```
ralfleibold: We might want to consider using unordered_map instead. I imagine this will be a bit faster as we do not need any ordering. This also applies to the std::map below.

Ozaq (changes requested):
Actual writing of the trace file should be done from python, see https://github.com/nico/ninjatracing/blob/main/ninjatracing for example. Then we can drop the C++ profiler / json dependency.

pitsteinbach:
OP#10616 
OP#10619

setup.py
```diff
@@ -253,7 +254,7 @@ def build_extension(self, ext: CMakeExtension) -> None:
     ext_modules=[CMakeExtension("python_bindings_jupedsim")],
     cmdclass={"build_ext": CMakeBuild},
     zip_safe=False,
-    python_requires=">=3.10,<3.14",
+    python_requires=">=3.10,<3.15",
```
Ozaq, line 256: This should be left out, has nothing to do with the PR.

python_modules/jupedsim/jupedsim/simulation.py
```diff
@@ -139,9 +139,16 @@ def __init__(
...
+        self._enable_timer_printout = False
         self._obj = py_jps.Simulation(
             model=py_jps_model, geometry=build_geometry(geometry)._obj, dt=dt
         )
+        self.push_timer("Total Simulation Time")
+
+    def __del__(self):
+        """Destructor for Simulation. Prints the timer output to the console."""
+        if self._enable_timer_printout:
+            self.print_timer()
```
Ozaq: Simulation should just provide access to the gathered data but not write it out. Writing and formatting is responsibility of the application using jupedsim. We may provide helper functionality to pretty print timings but this should something like:

```python
timing_str = jupedsim.pretty_print_timings(sim.timings)
# alternatively even more embedded in python stlye
print(sim.timings) # where whatever timings is implements __str__ and / or __repr__
```

python_modules/jupedsim/jupedsim/simulation.py
```diff
@@ -554,3 +561,45 @@ def switch_geometry(self, geometry: Geometry) -> None:
         """
         internal_geometry = build_geometry(geometry)
         self._obj.switch_geometry(internal_geometry._obj)
+    
+    def print_timer(self) -> None:
+        """Prints the timer entries to the console. By default only the main iteration loop is timed."""
+        self.pop_timer("Total Simulation Time")
```
Ozaq: If I understand the code correctly this will consume the pushed timer, which then in turn leads to different results when print_timer is called twice. If this is the case then the code is surprising callers. Printing should not change the state of what is being printed.

libsimulator/src/Simulation.cpp
```diff
@@ -59,34 +59,53 @@ const SimulationClock& Simulation::Clock() const
...
-    _perfStats.SetEnabled(status);
+    _perfStats.EnableProfiler(status);
 };
 
-PerfStats Simulation::GetLastStats() const
+const PerfStats& Simulation::GetLastStats() const
 {
-    return _perfStats;
+    return std::move(_perfStats);
 };
```
Ozaq: The move is a no-op since you now return a const ref. Why do you want to change to reference semantics here? The copy was intentional.

third-party/CMakeLists.txt
```diff
@@ -76,3 +76,14 @@ endif ()
 ################################################################################
 set(GLM_ENABLE_CXX_20 ON CACHE BOOL "" FORCE)
 add_subdirectory(glm)
+
+################################################################################
+# Chrome Trace Event Format Profiler
+################################################################################
+
+if (BUILD_PROFILER)
```
Ozaq: I would strongly prefer if we do not resort to conditional compilation unless there is no other way.

libsimulator/src/Tracing.cpp
```diff
@@ -3,29 +3,185 @@
...
+        // Probe already exists, return
+        return;
+    }
+}
+void PerfStats::PopProfilerProbe(const std::string& name)
+{
+    if(event_map.find(name) != event_map.end()) {
+        event_map[name]->stop();
+    }
+}
```
Ozaq: I did look a bit at the profiler lib implementation and the spec (https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU) and I am wondering why we bother here with the `DurationEvents` in the first place? (also I think we can relatively easy just implement the parts of the spec ourself, esp since the the profiling lib looks unmaintained anyways.

Think something like this:
```
tracing.RecordDurationBegin("foo");
// ... later
tracing.RecordDurationEnd();
```
Internally we check if tracing is enabled (backend present) and if yes we just write a simple event to a file or memory ringbuffer, e.g. based on something like this

```cpp
#include <fmt/std.h>
#include <chrono>
#include <cstdint>
#include <string_view>

#ifdef _WIN32
#include <windows.h>
using file_handle = HANDLE;
#else
#include <fcntl.h>
#include <unistd.h>
using file_handle = int;
#endif

file_handle open_trace(const char* path) {
#ifdef _WIN32
    return CreateFileA(path, GENERIC_WRITE, 0, nullptr,
                       CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr);
#else
    return open(path, O_WRONLY | O_CREAT | O_TRUNC, 0644);
#endif
}

void close_trace(file_handle fd) {
#ifdef _WIN32
    CloseHandle(fd);
#else
    close(fd);
#endif
}

void write_raw(file_handle fd, const char* data, size_t size) {
#ifdef _WIN32
    WriteFile(fd, data, (DWORD)size, nullptr, nullptr);
#else
    write(fd, data, size);
#endif
}

static uint64_t now_us() {
    return std::chrono::duration_cast<std::chrono::microseconds>(
        std::chrono::steady_clock::now().time_since_epoch()
    ).count();
}

void trace_begin(file_handle fd, std::string_view name) {
    char buf[256];
    auto result = fmt::format_to_n(buf, sizeof(buf),
        R"({{"ph":"B","name":"{:.64}","tid":{},"ts":{}}})" "\n",
        name, std::this_thread::get_id(), now_us());
    write_raw(fd, buf, result.size);
}

void trace_end(file_handle fd) {
    char buf[128];
    auto result = fmt::format_to_n(buf, sizeof(buf),
        R"({{"ph":"E","tid":{},"ts":{}}})" "\n",
        std::this_thread::get_id(), now_us());
    write_raw(fd, buf, result.size);
}
```

Ozaq (changes requested):

Ozaq:
> Actual writing of the trace file should be done from python, see https://github.com/nico/ninjatracing/blob/main/ninjatracing for example. Then we can drop the C++ profiler / json dependency.

@pitsteinbach I saw your comment in chat, I was thinking of essentially just keeping the events in memory and writing them out in python but that ofc comes with issues such as out of memory, lots of time spend in python. So yeah for the case when we want to collect traces writing them out from out native code is the right choice.

In general I believe we should drop the Profiler dependency and write the few event types we need ourselves, also keeping DurationEvents in a map to look them up is time spend needlessly because could just write an Duration End event to the event stream, End events are unnamed anyways according to spec.

ralfleibold:
We can remove the C-files of Perfetto. We only need the C++ ones..

We probably also need to mention somewhere that we use this as external lib as dependency?

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,62 @@
...
+    TimerEntry(const TimerEntry& other);
+    TimerEntry& operator=(const TimerEntry& other);
+    TimerEntry(TimerEntry&& other) noexcept;
+    TimerEntry& operator=(TimerEntry&& other) noexcept;
+    void start();
+    void stop();
+    uint64_t getDuration() const;
 };
 
 class PerfStats
```
ralfleibold: I think we need to disable copy constructor and assignment operator? We can consider move semantics, but not sure this is necessary.

pitsteinbach:
In an internal discussion about this PR, we discussed the following design changes:

- Have separate classes for timing and tracing purposes
- stick to copy semantics for the timer class, it is copied if needed to the corresponding python object
- use different log levels for the timer in order to more fine-grained control over which parts of the code are timed when.
- printing the timer happens in Python (`print(sim.timer)`), Python only get's the corresponding dictionary with the measured durations
- The tracing class should be usable in every part of the C++ code for debugging and hence needs to be contained in a singleton 
- The timer class optionally also triggers a profiling event
- The profiler can be enabled in Python 
- We would like to have scoped (event or timer is pushed upon creation and popped when exiting the scope) version of the timer and profiler

Potentially still up for debate, to we make the Timer class also a singleton and use a safe reference to it in Python, this would make it more flexible and potentially more natural to use.

libsimulator/src/Simulation.cpp
```diff
@@ -59,26 +59,51 @@ const SimulationClock& Simulation::Clock() const
...
+    }
+
+    {
+        JPS_SCOPED_TIMER(_timer, "Stage System", Detailed);
+        _stageSystem.Run(_stageManager, _neighborhoodSearch, *_geometry);
+    }
+
     {
-        auto t2 = _perfStats.TraceOperationalDecisionSystemRun();
+        JPS_SCOPED_TIMER(_timer, "Stategical Decision System", Detailed);
```
ralfleibold: Minor typo: "Strategical"

libsimulator/src/Simulation.cpp
```diff
@@ -59,26 +59,51 @@ const SimulationClock& Simulation::Clock() const
...
+        _stategicalDecisionSystem.Run(_journeys, _agents, _stageManager);
+    }
+
+    {
+        JPS_SCOPED_TIMER(_timer, "Tactical Decision System", Detailed);
```
ralfleibold: Personally I would put all 3 "main" steps ("Strategical", "Tactical" and "Operational") into "General"-scope. Especially routing is sth to be seen by default IMO.

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+#include <map>
+#include <memory>
+#include <optional>
+#include <string>
+#include <unordered_map>
+#include <utility>
+
+// Helper macro to create a scoped timer probe.
+// It creates a scope guard that starts the timer probe when it is created and stops the timer probe
+// when it goes out of scope. The name of the timer probe. The log level is used to filter which
```
ralfleibold: Is there something missing? "The name if the timer probe" is no real sentence. I think the input "name" is self-explaining - or make it a complete sentence.

libsimulator/src/Timing.cpp
```diff
@@ -0,0 +1,168 @@
...
+// Copy assignment operator.
+TimerEntry& TimerEntry::operator=(const TimerEntry& other)
+{
+    if(this != &other) {
+        started_at = other.started_at;
+        duration_in_microseconds = other.duration_in_microseconds;
+        running = other.running;
+    }
+    return *this;
+}
```
ralfleibold: Can you please remind me why we need the copy constructor? I still hope we can avoid it. If due to exposing to Python, I think we only need the name and duration, so maybe we can avoid?

libsimulator/src/Timing.cpp
```diff
@@ -0,0 +1,168 @@
...
+                   .count();
+    }
+    return duration_in_microseconds;
+}
+
+// Push a timer probe with the given name and log level.
+// If the log level of the timer probe is higher than the log level set in the Timer object,
+void Timer::pushTimerProbe(const std::string& name, int loglevel)
+{
+    if(loglevel > log_level) {
```
ralfleibold: This is really tough to read. Maybe we should call the input  `timer_probe_level` or the class-level one `max_log_level`?

libsimulator/src/Timing.cpp
```diff
@@ -0,0 +1,168 @@
...
+        return duration_in_microseconds +
+               std::chrono::duration_cast<std::chrono::microseconds>(
+                   std::chrono::high_resolution_clock::now() - started_at)
+                   .count();
+    }
+    return duration_in_microseconds;
+}
+
+// Push a timer probe with the given name and log level.
+// If the log level of the timer probe is higher than the log level set in the Timer object,
```
ralfleibold: This comment is truncated ("If" does not end).

libsimulator/src/Timing.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void Timer::popTimerProbe(const std::string& name)
+{
+    // use auto iter = timer_map.find(name) to avoid multiple lookups
+    auto iter = timer_map.find(name);
+    if(iter != timer_map.end()) {
+        iter->second.stop();
+    }
+    // if tracing is enabled a trace with the same name will also be poped from the
```
ralfleibold: This is technically not true, right? Tracing will just pop the *last* entry from the stack, or am I wrong?
Also minor typo: "poped" --> "popped"

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+        // Pointer to the Timer object that the guard will operate on.
+        // It is set to nullptr if the guard is moved from or if the log level of the timer probe is
+        // higher than the log level set in the Timer object, in which case the guard will not be
+        // active and will not record time.
+        Timer* stats{nullptr};
+        // Name of the timer probe that the guard will operate on.
+        std::string name;
+
+    public:
+        ScopedTimerProbeGuard() = default;
```
ralfleibold: Why do we need the default constructor?

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+        ScopedTimerProbeGuard(Timer& stats_ref, std::string timer_name, int loglevel)
```
ralfleibold: Better use `std::string_view`. Then you do not need to std::move below. We do not use perfect forwarding here anyways.

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+            }
+            return *this;
+        }
+    };
+
+    ~Timer() = default;
+    // Creates a scoped timer probe guard that starts the timer probe with the given name
+    // and log level when it is created and stops the timer probe when it goes out of scope.
+    [[nodiscard]] inline ScopedTimerProbeGuard
+    scopedTimerProbe(const std::string& name, int loglevel = 0)
```
ralfleibold: I suggest suing `std::string_view` here as well to avoid any copies. This also applies for other methods below.

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+    // Upon stopping the timer probe, the duration of the timer entry is updated with the time
+    // elapsed since it was started.
+    // If tracing is enabled a trace with the same name will also be poped from the
+    // ProfilerSingleton instance.
+    void popTimerProbe(const std::string& name);
+    // Returns the duration of the timer entry in microseconds. If the timer is still running, it
+    // returns the duration until now. if the timer entry does not exist, it returns 0.
+    uint64_t getTimerEntry(const std::string& name) const;
+    // Returns a formatted report of timer entries.
+    std::string printTimerEntries() const;
```
ralfleibold: We should rename as it does not print anymore.

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+    // Stops the timer probe with the given name.
+    // If the timer probe does not exist, this function does nothing.
+    // Upon stopping the timer probe, the duration of the timer entry is updated with the time
+    // elapsed since it was started.
+    // If tracing is enabled a trace with the same name will also be poped from the
+    // ProfilerSingleton instance.
+    void popTimerProbe(const std::string& name);
+    // Returns the duration of the timer entry in microseconds. If the timer is still running, it
+    // returns the duration until now. if the timer entry does not exist, it returns 0.
+    uint64_t getTimerEntry(const std::string& name) const;
```
ralfleibold: Rename, as it does not return the timer entry (anymore).

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+    // Returns a formatted report of timer entries.
+    std::string printTimerEntries() const;
+    // Returns a map of timer entry names to their durations in microseconds.
+    // If a timer is still running, it returns the duration until now.
+    // If a timer entry does not exist, it is not included in the map.
+    // PST: I choose a map here so that we always have the same order of entries when printing them.
+    std::map<std::string, uint64_t> getTimerEntries() const
```
ralfleibold: Similar: Rename to `GetDurations`?
Also maybe better `std::unordered_map`?

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,97 @@
...
+        ScopedProbeGuard& operator=(const ScopedProbeGuard&) = delete;
+        ScopedProbeGuard(ScopedProbeGuard&& other) noexcept : profiler(other.profiler)
+        {
+            other.profiler = nullptr;
+        }
+        ScopedProbeGuard& operator=(ScopedProbeGuard&& other) noexcept
+        {
+            if(this != &other) {
+                if(profiler) {
+                    profiler->popProbe();
```
ralfleibold: Why do we need to popProbe? This looks like an unexpected side effect. As the profiler is also still assigned below, it will pop again in destructor, right?

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,97 @@
...
+            }
+            return *this;
+        }
+    };
+
+    static ProfilerSingleton& instance() noexcept { return profiler; };
+
+    void enable();
+    void disable();
+    void pushProbe(const std::string& name);
```
ralfleibold: I suggest using `std::string_view` and also to define inline here for performance reasons.

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,97 @@
...
+    void popProbe();
+    [[nodiscard]] inline ScopedProbeGuard scopedProbe(const std::string& name)
+    {
+        return ScopedProbeGuard(this, name);
+    }
+
+    void dump(const std::string& filename);
```
ralfleibold: Maybe better be explicit in the method name. Maybe rename to `dump_and_reset`?

libsimulator/src/Simulation.hpp
```diff
@@ -85,6 +87,9 @@ class Simulation
     StageProxy Stage(BaseStage::ID stageId);
     CollisionGeometry Geo() const;
     void SwitchGeometry(std::unique_ptr<CollisionGeometry>&& geometry);
+    void PushTimer(const std::string& name);
+    void PopTimer(const std::string& name);
+    void SetLogLevelTimer(int level) { _timer.setLogLevel(level); };
```
ralfleibold: Better `SetTimerLogLevel`?

libsimulator/src/Simulation.hpp
```diff
@@ -85,6 +87,9 @@ class Simulation
     StageProxy Stage(BaseStage::ID stageId);
     CollisionGeometry Geo() const;
     void SwitchGeometry(std::unique_ptr<CollisionGeometry>&& geometry);
+    void PushTimer(const std::string& name);
```
ralfleibold: Shouldn't we have log-level here in the API as well?

python_bindings_jupedsim/trace.cpp
```diff
@@ -1,24 +1,31 @@
...
-        });
-}
+    py::class_<ProfilerSingleton>(m, "Trace")
+        .def_static(
+            "instance",
+            []() -> ProfilerSingleton& { return ProfilerSingleton::instance(); },
+            py::return_value_policy::reference)
+        .def("enable", [](ProfilerSingleton& ps) { ps.enable(); })
+        .def("disable", [](ProfilerSingleton& ps) { ps.disable(); })
+        .def("push_probe", [](ProfilerSingleton& ps, const std::string& name) { ps.pushProbe(name); })
```
ralfleibold: I do not think we should expose those to Python. I see tracing being used at very low-level and time-critical parts. I do not feel that Python-level is good here. What is your opinion?

python_modules/jupedsim/jupedsim/internal/tracing.py
```diff
@@ -3,36 +3,145 @@
...
+        self._prev_op_dec_time = 0
+        pass
 
-    @property
-    def iteration_duration(self) -> float:
-        """Time for one simulation iteration in us.
+    def __repr__(self) -> str:
+        """Prints the timer entries to the console. By default only the contributions with more than 0.01% are printed."""
+        # even if the timer has been stopped this should not change things
+        self._timer.pop_timer("Total Simulation Time")
```
ralfleibold: We start this in `simulation.py` and stop it here? This is not good. We should not do this. This has to be in one place. I think this all has to go into `simulation.py`? Not sure. We should discuss.

python_modules/jupedsim/jupedsim/internal/tracing.py
```diff
@@ -3,36 +3,145 @@
...
+
+    def enable(self) -> None:
+        """Enables the profiler."""
+        self._profiler.enable()
+
+    def disable(self) -> None:
+        """Disables the profiler."""
+        self._profiler.disable()
+
+    def dump(self, filename: str) -> None:
```
ralfleibold: Better: `dump_and_reset` as discussed in Tracing-class?

ralfleibold (changes requested):
Nice work albeit my many small comments. This is a great addition to our code base.

- Can we also add some really basic unit tests on Python level?
- You have added nice comments. Sometimes you commented on both the head nad the .cpp file. As this hard to keep in sync, just do in one location. (I prefer te header in this case.)

libsimulator/src/Timing.hpp
```diff
@@ -0,0 +1,164 @@
...
+// It also has a flag to indicate whether the timer is currently running or not.
+// The duration is summed, hence multiple calls to start and stop will accumulate the duration in
+// microseconds.
+class TimerEntry
+{
+    // Last start time of the timer entry.
+    std::chrono::high_resolution_clock::time_point started_at;
+    // Duration of the timer entry in microseconds.
+    // It is updated with the time elapsed since the last start time when the timer is stopped.
+    uint64_t duration_in_microseconds{0};
```
ralfleibold: Please define `duration_type` as uint64_t and use it here and in all code referring to duration. This way we can easily change the type if needed and it will expose accordingly just by recompiling.

libsimulator/src/Timing.cpp
pitsteinbach: removed as I restrict function comments  to the header files

libsimulator/src/Timing.cpp
pitsteinbach: You are right, I forgot to change this comment after changing the profiling library.

libsimulator/src/Timing.hpp
pitsteinbach: see comment, I think the regular map makes more sense, here as I personally and probably others would like to have an ordered view of the timer entry strings. I could also reorder in python but since this happens after simulation it is not time critical.

libsimulator/src/Simulation.hpp
pitsteinbach: My thinking was that if you push a timer via the API it has level 0 (the default), and is pushed always since the user explicitly added this call.

python_modules/jupedsim/jupedsim/internal/tracing.py
pitsteinbach: fixed, in the constructor of the Python Timer object the timer is started and then stopped upon printing

ralfleibold (dismissed):
Nice work. Thanks for your patience.

third-party/CMakeLists.txt
```diff
@@ -76,3 +76,22 @@ endif ()
...
+set_target_properties(perfetto PROPERTIES POSITION_INDEPENDENT_CODE ON)
+
+if (WIN32)
+  # The perfetto library contains many symbols, so it needs the big object
+  # format.
+  target_compile_options(perfetto PRIVATE "/bigobj")
+  # Winsock symbols (e.g. accept) are provided by ws2_32 on Windows.
+  target_link_libraries(perfetto PUBLIC ws2_32)
+  # Disable legacy features in windows.h.
+  add_definitions(-DWIN32_LEAN_AND_MEAN -DNOMINMAX)
```
Ozaq: This will set compile definitions for all files. we only want to set this for compilation of perfetto.cc
```suggestion
    target_compile_definitions(
        perfetto 
        PRIVATE 
            WIN32_LEAN_AND_MEAN 
            NOMINMAX
    )
```

third-party/CMakeLists.txt
```diff
@@ -76,3 +76,22 @@ endif ()
...
+endif (WIN32)
```
Ozaq, line 102: indentation 2 instead of 4

third-party/perfetto/perfetto.cc
Ozaq, line 1: Please add a readme specifying the version used, the files seem not to contain any version identifier.

python_modules/jupedsim/tests/test_timer_profiler.py
```diff
@@ -0,0 +1,84 @@
+import time
+
+import pytest
+from jupedsim.models.collision_free_speed import CollisionFreeSpeedModel
+
+
+def test_timer_integration_small_simulation(tmp_path):
+    """
+    Integration test: exercise the real C++ Timer API exposed via jupedsim.native.
+    This test will be skipped if the extension or the expected methods are not available.
```
Ozaq: I don't see any conditional execution of this test. Also i think we should not conditionally run tests if possible.

python_modules/jupedsim/tests/test_timer_profiler.py
```diff
@@ -0,0 +1,84 @@
...
+        "enable",
+        "disable",
+        "push_probe",
+        "pop_probe",
+        "dump_and_reset",
+    ]
+    if not all(hasattr(profiler, m) for m in required):
+        pytest.fail(
+            "py_jps.Profiler missing required methods for integration test"
+        )
```
Ozaq: Why are you testing this? You are calling the methods immediately after anyways.

python_modules/jupedsim/tests/test_timer_profiler.py
```diff
@@ -0,0 +1,84 @@
...
+    geom = [(0.0, 0.0), (1.0, 0.0), (1.0, 1.0), (0.0, 1.0)]
+    sim = Simulation(model=CollisionFreeSpeedModel(), geometry=geom, dt=0.01)
+
+    # underlying C++ simulation object exposed as _obj
+    timer = sim.get_last_timer()
+
+    # Ensure the pybind Simulation exposes timer methods
+    assert hasattr(timer, "push_timer")
+    assert hasattr(timer, "pop_timer")
+    assert hasattr(timer, "elapsed_time_us")
```
Ozaq: Why are you testing this? You are calling the methods immediately after anyways.

python_modules/jupedsim/jupedsim/internal/tracing.py
```diff
@@ -3,36 +3,135 @@
...
+        Args:
+            name: Name of the probe to be pushed.
+        """
+        self._profiler.push_probe(name)
 
-    def __str__(self) -> str:
-        return self._obj.__repr__()
+    def pop_probe(self) -> None:
+        """Pops the last pushed probe. The probe will be stopped and the elapsed time will be recorded."""
+        self._profiler.pop_probe()
```
Ozaq: This should better be just a bunch of functions since its using the singleton anyways. The C++ Singelton can be exposed as global python object to save on crossing the language barrier if that is a concern.

This in turn would allow with a few lines of code to add a function based context manager

```python
from contextlib import contextmanager

@contextmanager
def trace_region(name):
    push_probe(name)
    try:
        yield
    finally:
        pop_probe()

# Then we wrap code in a trace event like this:
with trace_region("name"):
    # ... code

# To wrap whole functions we can provide a decorator such as:
import functools

def trace_event(fn):
    @functools.wraps(fn)
    def wrapper(*args, **kwargs):
        push_probe(f"{fn.__name__}()")
        try:
            return fn(*args, **kwargs)
        finally:
            pop_probe()
    return wrapper

# then just slap the decorator on your function decl
@trace_event
def foo():
    pass
```

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,104 @@
...
+            other.profiler = nullptr;
+        }
+        ScopedProbeGuard& operator=(ScopedProbeGuard&& other) noexcept
+        {
+            if(this != &other) {
+                profiler = other.profiler;
+                other.profiler = nullptr;
+            }
+            return *this;
+        }
```
Ozaq: Semantically they should not be movable. Right now this type tells me from ts name that its ensuring the lifetime of a probe is tied to a scope. In this scenario I cannot see it being useful that the type is movable. Am I missing something?

libsimulator/src/Tracing.hpp
```diff
@@ -1,41 +1,104 @@
...
+            if(this->profiler) {
+                this->profiler->pushProbe(name);
+            }
+        }
+        ~ScopedProbeGuard()
+        {
+            if(profiler) {
+                profiler->popProbe();
+            }
+        }
```
Ozaq: Is there a specific reason why you take the indirection of a factory method + backpointer instead of directly constructing `ScopedProbeGuard` and using the singleton directly? 

E.g.
```
ScopedProbeGuard(std::string_view name) {
    ProfilerSingelton.Instance().pushProbe(name);
}

~ScopedProbeGuard() {
    ProfilerSingelton.Instance().popProbe();
}

```

python_modules/jupedsim/jupedsim/native/__init__.py
```diff
@@ -6,4 +6,8 @@
     from .. import py_jupedsim as py_jps
     from ..py_jupedsim import *  # noqa: F403
 
-__all__ = [name for name in dir(py_jps) if not name.startswith("__")]
+__all__ = [
+    name
+    for name in dir(py_jps)
+    if not name.startswith("__") or name in ["Timer"]
```
Ozaq: Why is this required?

libsimulator/src/Tracing.cpp
```diff
@@ -1,31 +1,97 @@
...
+    if(output.is_open()) {
+        output.write(trace_data.data(), static_cast<std::streamsize>(trace_data.size()));
+        output.flush();
     } else {
-        return std::nullopt;
+        std::cerr << "Failed to open Perfetto output file: " << filename << std::endl;
+    }
+
+    tracing_session.reset();
+}
```
Ozaq: Perfetto can periodically write its in memory buffer, there is not need to write this ourselves IMO. 

https://github.com/google/perfetto/blob/facb3b402650e868128127b02aac5d2403da380a/docs/concepts/config.md?plain=1#L235

Ozaq (commented):
Ill leave it to you to merge this or not but I don't think this should be merged in its current form.

What would help the PR is to split out timers from probes.

Further Perfetto SDK already provides RAII style probes and functionality to not record events. I think we should not reimplement them. Sure we need to write some init function but after that I would favor using the Perfetto SDK macros to create probes. This also forces us to use static strings which gives us string interning. Our current implementation adds a heap allocation on every trace event (perfetto::DynamicString) and hence cannot intern strings vastly making the trace events larger.

python_modules/jupedsim/jupedsim/native/__init__.py
pitsteinbach: I initially got some errors importing the Python Timer class. Since there is no Timer class in `jupedsim.internal` that is created by the python bindings, I thought that the error came from this line. However, it seems fine without.

libsimulator/src/Tracing.hpp
pitsteinbach: I removed them.

libsimulator/src/Tracing.hpp
pitsteinbach: I changed to a variation of your suggestions. I don't really understood why that is an issue, if you could send me a longer explanation on our internal chat that would be great.

python_modules/jupedsim/tests/test_timer_profiler.py
pitsteinbach: I removed them

libsimulator/src/Tracing.cpp
pitsteinbach: Are you sure that this works as well when using the SDK, your link is from the CLI.
Their example when using the SDK is performing the same steps:
https://github.com/google/perfetto/blob/main/examples/sdk/example.cc
Additionally, if you want to set the name of the output on dump this is the way to go.
When dumping regularly you need to setup the file buffer beforehand, which in turn make the use of a more sophisticated constructor or `enable` function necessary.

python_modules/jupedsim/jupedsim/internal/tracing.py
pitsteinbach: The decorators are a good idea, I implemented one function that can be used either as decorator or as context manager and I call both of them `trace_event`.
I also changed the `grosser_stern.py` test to now use this tracing option for parts of the Python code.
Screenshot 2026-05-08 at 13 49 40

pitsteinbach:
I separated the Timer and Profiler now in terms of classes.
Additionally, I am now using the macro form Perfetto that already allowes recording scopes, hence we are not duplicating functionality there.

However, I like the convenience of using a single macro for timing and tracing of the most common parts of jupedsim. Therefore, the macro call to the timer now directly also triggers the tracing. 

Within Python profiler is now a global variable and we interact with it using functions. To trace functions or regions of code in Python. I added a decorator and context manager that pushes and pops using the C++ singleton, as suggested by @Ozaq. 

In the same spirit, timer now also can be used with a decorator or context manager to record the time of Python code regions.

Tests have also been updated.

ralfleibold:
Looks great now. All issues addressed, the final version is faster and does use static strings for tracing whenever possible.

ralfleibold (approved):

libsimulator/src/Tracing.cpp
pitsteinbach: We will go forward with the current solution if we run into problems or figure out more about Perfetto we will have another look at his feature and try to figure out how CLI addresses this feature. Currently not a priority and Perfetto SDK docs is unfortunately not very extensive.

PR #1559, opened 2026-04-27
Backport warp-driver model to 1.x

ralfleibold:
Backport the WarpDriver model from feature branch based on main to 1.4 release branch to be integrated into an upcoming 1.4 release.


libjupedsim/include/jupedsim/warp_driver_model.h
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * Creates a WarpDriverModel builder.
+ * @param timeHorizon [Mohcine]...
+ * @param stepSize [Mohcine]...
+ * @param sigma [Mohcine]...
+ * @param timeUncertainty [Mohcine]...
+ * @param velocityUncertaintyX [Mohcine]...
+ * @param velocityUncertaintyY [Mohcine]...
+ * @param numSamples [Mohcine]...
+ * @param rngSeed [Mohcine]...
```
ralfleibold: @chraibi Can you please document the parameters? If we do not want to expose all of them to the C-API, we should remove accordingly (if some have mandatory defaults).

examples/benchmark_models/benchmark_models.py
```diff
@@ -0,0 +1,648 @@
...
+        ax.set_aspect("equal")
+
+
+def compute_fluctuation(sqlite_file):
+    """Compute velocity fluctuation time series from a trajectory file."""
+    traj = pedpy.load_trajectory_from_jupedsim_sqlite(Path(sqlite_file))
+    fps = traj.frame_rate
+    dt = 1.0 / fps
+
+    individual_velocity = pedpy.compute_individual_speed(
```
ralfleibold: I assume this to be the same formula as in `fluctuation.py`. We probably want to share the code. If this is generally good to have, we could push this to the Python-API. If only useful for benchmarking, we should leave this here, but still share the function.

examples/benchmark_models/benchmark_models.py
```diff
@@ -0,0 +1,648 @@
...
+        color = model_colors[model_name]
+        ax.plot(time_s, fluctuation, color=color, alpha=0.7, linewidth=0.5)
+        ax.set_xlabel("Time (s)", fontsize=8)
+        ax.set_ylabel("$\\phi_v$ (m/s²)", fontsize=8)
+        ax.set_title(model_name, fontsize=9)
+        ax.set_ylim(bottom=0)
+        ax.grid(True, linestyle="--", alpha=0.5)
+
+
+def compute_order_parameter(sqlite_file):
```
ralfleibold: Same comment as with `compute_fluctuation`: We should share this utility function (here: with `compute_order_parameters.py`).

libsimulator/src/WarpDriverModel.cpp
```diff
@@ -0,0 +1,706 @@
...
+// ============================================================================
+
+void WarpDriverModel::IntrinsicField::Compute(double sigma)
+{
+    nx = static_cast<int>(std::round((xMax - xMin) / dx)) + 1;
+    ny = static_cast<int>(std::round((yMax - yMin) / dy)) + 1;
+    values.resize(static_cast<size_t>(nx * ny), 0.0);
+    gradients.resize(static_cast<size_t>(nx * ny), Point{0.0, 0.0});
+
+    const double sigma2 = sigma * sigma;
```
ralfleibold: maybe better `sigma_squared`

libsimulator/src/WarpDriverModel.cpp
```diff
@@ -0,0 +1,706 @@
...
+
+    // Compute I(x,y) = (f * g)(x,y) where g = unit disk, f = Gaussian(sigma).
+    // For each grid point, numerically integrate the convolution over the disk.
+    const double intStep = 0.05;
+    const double intRadius = 1.0; // unit disk support
```
ralfleibold: What does "int" stand for here: "internal"? This is confusing with the "int" type as those are double-values. Can we please rename?

python_bindings_jupedsim/warp_driver_model.cpp
```diff
@@ -0,0 +1,168 @@
...
+            })
+        .def_property(
+            "desired_speed",
+            [](const JPS_WarpDriverModelState_Wrapper& w) {
+                return JPS_WarpDriverModelState_GetV0(w.handle);
+            },
+            [](JPS_WarpDriverModelState_Wrapper& w, double v0) {
+                JPS_WarpDriverModelState_SetV0(w.handle, v0);
+        //     })
+        // .def_property(
```
ralfleibold: @chraibi It is not clear to me whether we expose all the parameters read-write to the user.

libjupedsim/include/jupedsim/warp_driver_model.h
chraibi: done

examples/benchmark_models/benchmark_models.py
chraibi: done

libsimulator/src/WarpDriverModel.cpp
chraibi: done

python_bindings_jupedsim/warp_driver_model.cpp
chraibi: We don't expose all parameters. I just expose what the user might need. The internal parameters, are better kept internal

ralfleibold:
OP#10610

PR #1573, opened 2026-04-29
Hotfix for 1.4.0 release due to missing C-header in install

ralfleibold:
warp_driver_model.h is not installed via "cmake --install". This is purely an issue when using JuPedSim via C-API and does not affect the Python module at all.


ralfleibold:
Wrong branch picked - should have been rls-1.4, not master. Closing.

Issue #1575, opened 2026-04-30
Index out of bounds access in RoutingEngine


ralfleibold:
Discussion https://github.com/PedestrianDynamics/jupedsim/discussions/1520 revealed an issue in RoutingEngine.cpp where an index is accessed beyound the size of the `std::vector`. The issue is that the loop runs until `<= size of the vector` and then accesses this element. The loop is simply running 1 time too often.

In addition there is some dead `if` code at the start of that loop which can be removed along the way.

### Important note:
This issue - according to my analysis - did **not** affect the routing outcome. While indeed some memory was addressed outside the boundary of the vector, the memory was not used in active code as some specific instruction was done afterwards to run certain code if the index was indeed outside the range.

ralfleibold:
Fixed by https://github.com/PedestrianDynamics/jupedsim/pull/1576.

PR #1576, opened 2026-04-30
Fixes 1575: Index out of bounds issue in RoutingEngine

ralfleibold:
In RoutingEngine:
- straightenPath runs a for-loop from 1 to size of the path vector (including that item) and accesses each element. This obviously accesses an item outside the defined vector.
- There is some dead code directly afterwards which compares `face_from->neighbor(0) == face_to` but does nothing in the body. Just checking neighbor 0 looks incomplete as well. The code should be removed.

This addresses #1575.


pitsteinbach (approved):
looks good to me, potentially we do one iteration too much as mentioned, but keep it for consistency.

PR #1578, opened 2026-05-04
Backport HDF5 trajectory writer to rls-1.4 branch.

ralfleibold:
This is based on PR https://github.com/PedestrianDynamics/jupedsim/pull/1564. 

As a side effect we will also get the doc how to write a custom serialization to rls-1.4 branch.

### DIFF
--- a/examples/example_hdf5_writer.py
+++ b/examples/example_hdf5_writer.py
@@ -87,7 +87,6 @@ def main() -> None:
 
     fig, ax = plt.subplots(figsize=(8, 6))
     pedpy.plot_trajectories(traj=traj, walkable_area=walkable_area, axes=ax)
-    print(walkable_area)
     ax.set_aspect("equal")
     ax.set_title("Hdf5TrajectoryWriter -> pedpy.plot_trajectories")
     png = out.with_suffix(".png")
--- a/examples/example_hdf5_writer.py
+++ b/examples/example_hdf5_writer.py
@@ -1,0 +1,99 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+"""Example: HDF5 trajectory writer.
+
+Runs a short simulation, writes the trajectory to an HDF5 file in a
+layout compatible with the loaders provided by `PedPy
+<https://github.com/PedestrianDynamics/PedPy>`_ for the Pedestrian
+Dynamics Data Archive format, then reads the file back with
+PedPy and plots the trajectories on top of the walkable area.
+
+Extra dependencies for this example (beyond JuPedSim itself):
+
+    pip install h5py pedpy matplotlib
+
+The script writes ``example_traj.h5`` and ``example_traj.png`` to the
+current working directory.
+"""
+
+import pathlib
+import sys
+
+import jupedsim as jps
+import matplotlib.pyplot as plt
+import pedpy
+from shapely import GeometryCollection, Polygon
+
+
+def main() -> None:
+    if jps.Hdf5TrajectoryWriter is None:
+        sys.exit("h5py is not installed. Install with: pip install h5py")
+
+    room1 = Polygon([(0, 0), (10, 0), (10, 10), (0, 10)])
+    room2 = Polygon([(15, 0), (25, 0), (25, 10), (15, 10)])
+    corridor = Polygon([(10, 4.5), (28, 4.5), (28, 5.5), (10, 5.5)])
+
+    area = GeometryCollection(corridor.union(room1.union(room2)))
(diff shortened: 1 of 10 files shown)


examples/example_hdf5_writer.py
```diff
@@ -0,0 +1,99 @@
...
+        trajectory_file=out
+    )
+    print(
+        f"Loaded {len(traj.data)} rows, {traj.data['id'].nunique()} agents, "
+        f"fps={traj.frame_rate}"
+    )
+
+    fig, ax = plt.subplots(figsize=(8, 6))
+    pedpy.plot_trajectories(traj=traj, walkable_area=walkable_area, axes=ax)
+    print(walkable_area)
```
ralfleibold: I don't think we should output this.

python_modules/jupedsim/jupedsim/hdf5_serialization.py
```diff
@@ -0,0 +1,406 @@
...
+        give only a few percent extra savings at noticeable extra
+        write cost.
+    """
+
+    def __init__(
+        self,
+        *,
+        output_file: pathlib.Path,
+        every_nth_frame: int = 4,
+        commit_every_nth_write: int = 100,
```
ralfleibold, line 119: IMO this is a pretty large number. Shall we really use `100` as default? Is this based on any experiments?

python_modules/jupedsim/jupedsim/hdf5_serialization.py
```diff
@@ -0,0 +1,406 @@
...
+        self._xmin = float("inf")
+        self._xmax = float("-inf")
+        self._ymin = float("inf")
+        self._ymax = float("-inf")
+
+    def begin_writing(self, simulation: Simulation) -> None:
+        if self._file is None:
+            raise TrajectoryWriter.Exception("File already closed.")
+
+        fps = 1.0 / simulation.delta_time() / self._every_nth_frame
```
ralfleibold, line 167: Is this correct? IMO we do not reduce the frames per second by a larger value of "every nth frame", do we?

systemtest/test_hdf5_writer.py
```diff
@@ -0,0 +1,132 @@
...
+
+h5py = pytest.importorskip("h5py")
+
+
+@pytest.fixture
+def square_simulation(tmp_path: pathlib.Path):
+    area = GeometryCollection(
+        shapely.Polygon([(0, 0), (10, 0), (10, 10), (0, 10)])
+    )
+    out = tmp_path / "traj.h5"
```
ralfleibold: Better: `ouput_filename`

python_modules/jupedsim/jupedsim/hdf5_serialization.py
```diff
@@ -0,0 +1,406 @@
...
+        )
+
+        self._file.attrs["schema_version"] = SCHEMA_VERSION
+        self._file.attrs["producer"] = "JuPedSim"
+        try:
+            from jupedsim import __version__ as _jps_version
+
+            self._file.attrs["producer_version"] = _jps_version
+        except Exception:
+            pass
```
ralfleibold, line 220: Is this optional? Or should we set sth like "unknown" here?

ralfleibold (commented):
Looks good. Some general comments and some clarifications.

python_modules/jupedsim/jupedsim/hdf5_serialization.py
ralfleibold: Revisited the logic: We write every nth frame but still enumerate the frames with 1, 2, 3... This mans the frames-per-second indeed decreases with a higher value of `every_nth_frame`.

examples/example_hdf5_writer.py
ralfleibold: Removed

python_modules/jupedsim/jupedsim/hdf5_serialization.py
ralfleibold: It was decided to keep this as-is for now. User can overwrite the setting. With more experience, we might change this default value in the future.

systemtest/test_hdf5_writer.py
ralfleibold: Changed.

python_modules/jupedsim/jupedsim/hdf5_serialization.py
ralfleibold: After discussion, we keep this as-is. We should always have a version - and otherwise the "producer_version" should be fine not being set.

PR #1580, opened 2026-05-08
Fix routing issues on master branch

ralfleibold:
There are actually 2 issues:

1. The loop in straightenPath() runs one item too short. 
2. The Python interface did not accept a pair of doubles anymore.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1580/


pitsteinbach (dismissed):
Looks good to me.

pitsteinbach:
It would be great if you can add the test that helped you catch the bug so our CI catches this in the future.

ralfleibold:
Thanks, I have added the test.

PR #1583, opened 2026-05-08
Fix python tests and add tests from python_modules/jupedsim/tests to CI

ralfleibold:
Not all Python tests are run during CI. Therefore one needs to add the ones from python_modules/jupedsim/tests to the tests being executed with cmake test target. As the name was "systemtests", rename this to "pythontests" to have a more common name.

There are some tests that either fail or are not properly coded. Especially:
- systemtest/test_dxf_parser.py
  It does not base all files being accessed relative to repository root. Doing so lets the tests fail due to missing imports. As a first shot we will skip this test if not all necessary modules are available. We probably want to fix this later by maybe having tests with additional requirements.
- python_modules/jupedsim/tests/test_build_geometry.py
  The import is wrong "jupedsim.util" --> "jupedsim.geometry_utils".
- python_modules/jupedsim/tests/test_distributions.py::test_distribution_till_full_creates_correct_points
  The last parameterized input leads to some failure expecting 44 samples but getting just 41.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1583/


pitsteinbach (approved):
looks good to me, nice that the tests are running regularly again!

PR #1584, opened 2026-05-11
Correct function argument for Timer context manager

pitsteinbach:
During the previous PR, I added context manager and decorators to the timer class. 
Unfortunately, I  missed changing back some parameters in the context manager function call.

I changed them accordingly, the unit test in place would have caught it.
We should move forward with #1583, so that the CI catches these silly mistakes in the future.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1584/


ralfleibold:
Thanks for the quick fix.

ralfleibold (approved):

PR #1585, opened 2026-05-11
Examples of bad A* routing: Shortest paths are longer than expected.

ralfleibold:
The current A* algorithm used for shortest path calculattion has some issues in certain situations. Examples are added to show-case these issues and to be able to compare this in the future with better routing algorithms.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1585/

PR #1586, opened 2026-05-11
Fix wrong CDT segment lookup in A* successor expansion

chraibi:
`cdt.segment(target, idx)` was returning an arbitrary edge of `target` instead of the shared edge between the current face and its neighbor. `idx` is the neighbor index in the CURRENT face, but CGAL's per-face vertex indexing is not symmetric across neighbors, so the segment must be queried from the current face.

The wrong edge produced bogus g/h values for some successors, causing A* to consistently mis-rank channels and return long detours even when a near-direct channel existed. 

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1586/


chraibi:
pr_buw_after
pr_buw_before

chraibi:
pr_aknz_evac_after
pr_aknz_evac_before

chraibi:
The examples added in #1585 are used to test this PR.

ralfleibold:
Thanks. This looks great. We should backport this fix to 1.4 Branch.

ralfleibold (dismissed):

PR #1588, opened 2026-05-19
Fix another issue in A* with last triangle and add a test for it.

ralfleibold:
This is cheap to fix for the last triangle but can (and will) also happen for intermediate triangles with A*. The latter would be insanely expensive and defeat way the A* algorithm works. This has to be fixed by different routing algorithm.

Basically `straightenPath()` is called each time you hit the final destination triangle to ensure getting a shorter path if the funneling algorithm yields it.

A corresponding routing test is added that passes only with this fix.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1588/

### DIFF
--- a/systemtest/test_routing.py
+++ b/systemtest/test_routing.py
@@ -59,6 +59,15 @@ BAD_ASTAR_ROUTINGS = [
         "max_diff": 0.2,
     },
     {
+        "test_name": "corner_with_shortcut2",
+        "description": "Same starting point, but end points differ just 0.03 on y axis, but total distance diff was >>0.03",
+        "wkt_path": "examples/geometry/corner_with_shortcut.wkt",
+        "error_type": "max_diff",
+        "path1": [(11.80, 1.00), (28.50, 13.54)],
+        "path2": [(11.80, 1.00), (28.50, 13.55)],
+        "max_diff": 0.01,
+    },
+    {
         "test_name": "aknz_evac",
         "description": "Direct path possible, but path was ways longer",
         "error_type": "direct path possible",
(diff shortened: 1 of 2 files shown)


systemtest/test_routing.py
```diff
@@ -106,6 +115,9 @@ def test_max_diff_example(test_entry):
 
     distance1 = path_distance(path1)
     distance2 = path_distance(path2)
+    print(
```
chraibi: Maybe remove this print

systemtest/test_routing.py
```diff
@@ -58,6 +58,15 @@ def test_routing_engine_without_excluded_areas():
...
     },
+    {
+        "test_name": "corner_with_shortcut2",
+        "description": "Same starting point, but end points differ just 0.03 on y axis, but total distance diff was >>0.03",
+        "wkt_path": "examples/geometry/corner_with_shortcut.wkt",
+        "error_type": "max_diff",
+        "path1": [(11.80, 1.00), (28.50, 13.54)],
+        "path2": [(11.80, 1.00), (28.50, 13.55)],
+        "max_diff": 0.01,
+    },
```
chraibi, line 69: Does this test fail before this PR?

chraibi (dismissed):

systemtest/test_routing.py
ralfleibold: Yes, it does.

systemtest/test_routing.py
ralfleibold: Done. Thanks!

chraibi (approved):

PR #1589, opened 2026-05-19
Speed-up of A* implementation

ralfleibold:
- Optimize memory management in A* implementation by using a `std::deque` instead of a `std:vector` as states only grow at the end.
- Do not use `std::shared_ptr` as the reference counting is not needed. Instead store all states in the `std::deque` (--> ownership) and only use the direct pointers in open and closed states.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1589/

### DIFF
--- a/libsimulator/src/RoutingEngine.cpp
+++ b/libsimulator/src/RoutingEngine.cpp
@@ -16,9 +16,9 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <deque>
 #include <limits>
 #include <map>
-#include <memory>
 #include <unordered_map>
 #include <utility>
 #include <vector>
@@ -202,7 +202,7 @@ std::vector<Point> RoutingEngine::ComputeAllWaypoints(Point currentPosition, Poi
                     // Unlike in A* this is only a first candidate solution
                     // Now compute the actual path length via funnel algorithm
                     // store path and length if this variant is the shortest found so far
-                    const SearchState dest_state{g_value, h_value, to, current_state.get()};
+                    const SearchState dest_state{g_value, h_value, to, current_state};
                     const auto vertex_ids = dest_state.path();
                     const auto found_path =
                         straightenPath(currentPosition, destination, vertex_ids);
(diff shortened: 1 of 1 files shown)


ralfleibold:
Several optimizations are done:
- Optimize memory management: Do not use `std::shared_ptr` but store all search states inside a `std::deque` which has the property of not changing addresses to its elements. Therefore use the direct pointers to the `SearchState` instead of the shared pointers.
- Each A* search iteration called `std::make_heap`. C++ has functions to keep the heap intact (`push_heap` / `pop_heap`). This reduces quite a bit of unnecessary overhead.
- `closed_states` grows and each search iteration, there are `find`-operations done inside that container. While the current `std::map` was already helpful here, an `std::unordered_map` has _typically_ better properties for inserting/searching/deleting elements.

The picture below shows the impact of each of these optimizations for an example with sufficiently large number of triangles (geometry in examples/geometry/SiB2023_entrance_jupedsim.wkt). Partially the time spent in routing is cut in half. (Measurements done on AMD processor using Linux, JuPedSim compiled with gcc-13.)

benchmark_SiB2023_entrance_jupedsim

libsimulator/src/RoutingEngine.cpp
pitsteinbach:
```suggestion
            // underestimation of the path length the exact path cannot be shorter than what we have
```

pitsteinbach:
```suggestion
            // The h-value is the distance between the goal and the closest point on the edge
```

pitsteinbach:
```suggestion
            //  This is an underestimate because the Euclidean distance metric used for the
```

pitsteinbach:
```suggestion
            // and emits a warning when capturing 'target' directly As of C++20 this SHOULD(TM)
```

pitsteinbach:
```suggestion
    // This is an overestimation but IMO preferable to repeated allocations.
```

pitsteinbach (dismissed):
Nice work, I can confirm that this makes it faster on Mac as well.
While looking at the code, I found some typos that I added as suggestions.

libsimulator/src/RoutingEngine.cpp
ralfleibold: Done.

ralfleibold: Done.

ralfleibold: Done.

ralfleibold: Done.

ralfleibold: Done.

PR #1593, opened 2026-05-26
Reorganise the operational models 

pitsteinbach:
Since, we have an ever increasing number of Operational Models, I thought it would be good to reorganise these a bit and collect all Operational Models in a subfolder, currently also regrouping per model.

I updated the CMake structure such that every directory has `CMakeLists.txt` that collects all source files in an array, in addition also the include directories are collected.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1593/


ralfleibold:
Reorganizing the dir structure is a good call. Your approach works fine, but I think there is a slightly superior patter with cmake.

### Suggested Change
- I am not sure whether to create a new CMakeLists.txt file in src. But not too strong of an opinion here. But I would define in the libsimulator/CMakeLists.txt the following:
```cmake
add_library(simulator STATIC)  # <-- no sources here anymore

# Non-model sources
target_sources(simulator PRIVATE
    src/AABB.cpp
    src/AABB.hpp
    ...
 )
 
# Operational models
add_subdirectory(src/OperationalModels)

...

target_include_directories(simulator PUBLIC
    ${CMAKE_CURRENT_SOURCE_DIR}/src
    ${CMAKE_CURRENT_SOURCE_DIR}/src/OperationalModels   # <-- to be added
)
```

- Then in libsimulator/src/OperationalModels/CMakeLists.txt:
```cmake
target_sources(simulator PRIVATE
    OperationalDecisionSystem.hpp
    OperationalModel.hpp
    OperationalModelUpdate.hpp
    OperationalModelType.hpp
)

add_subdirectory(AnticipationVelocityModel)
add_subdirectory(CollisionFreeSpeedModel)
...
```

- And e.g. libsimulator/src/OperationalModels/CollisionFreeSpeedModel/CMakeLists.txt:
```cmake
target_sources(simulator PRIVATE
    CollisionFreeSpeedModel.cpp
    CollisionFreeSpeedModel.hpp
    CollisionFreeSpeedModelBuilder.cpp
    CollisionFreeSpeedModelBuilder.hpp
    CollisionFreeSpeedModelData.hpp
    CollisionFreeSpeedModelUpdate.hpp
)
```

### Some notes
- IMO this looks a bit leaner and makes it easier to add new models.
- I requires some change in the source files: We need to prefix the "include" of the .hpp files with their model dir, e.g.
  ```c++
  #include "CollisionFreeSpeedModel/CollisionFreeSpeedModel.hpp"
   ```
   This makes includes a bit more verbose, but I think it is worth it.
- I am unsure whether the dir should be "OperationalModels" (with an 's' at the end) or "OperationModel". One the one hand it contains all operational models, but we also have the operational decision system in here - and therefore contains more than just the models. I feel either we only move the models and then add the "s" or we push all operational model related parts in the subdir and remove the "s". But only a slight preference. :smiley: 

What do you think?

ralfleibold (changes requested):
Oops - forgot to use the official "Submit Review". Please see my previous comment.

pitsteinbach:
Thanks for the input, indeed using the newer target based source collection seems like a clean and efficient way to move forward. I adopted your proposition with the alteration that I also include the model directories in the include paths, this avoids the long include statements.
Also I moved the `OperationalDecisionSystem.hpp` again to the base src directory to justify the "s" in the naming 😄

ralfleibold (approved):
Looks great now. Thanks!

PR #1594, opened 2026-05-29
Store pointer in Neighbourhood Search grid instead of Agents

pitsteinbach:
Currently, the Neighbourhood Search stores all agents on a 2D-grid.
This is done by storing a copy of the Agent in a vector per grid point.
Essentially, this means that all agents are stored 2 times: once in the overall vector containing the agents and once on the grid inside Neighbourhood search.
If a model or another function requests all agents within a certain range, the agents are copied again into a new vector that is returned.

In this PR, I propose to store not the agent itself on the grid but rather a pointer to that agent. This removes the additional copy of alle agents in the neighbourhood system. 
To make sure that theses references are valid during an iteration cycle, the container holding the agents is changed from a `std::vector` to a `std::deque`. In general, the container holding the agents is now defined in `GenericAgent.hpp` and used throughout the codebase. 

I also removed the `AddAgent` call to the neighbourhood system, since the neighbourhood system is updated in the iteration cycle anyhow. One possibility, where the neighbourhood system would be empty is if `AgentsInPolygon` or `AgentsInRange` is called, therefore a grid update is performed prior to getting the neighbours. Potentially, a state variable could be useful if these functions are called often.

I also added a performance test with 15k agents in a large geometry, as well as a notebook that can be used to draw box plots for the timing results of various program components.

Optionally, I also added a function that gets the pointers of the agents in a certain range instead of a copy of the agent, I also include it in my benchmark comparison.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1594/


pitsteinbach:
Timing comparison for the changes implemented in this PR compared to the master branch.

30-35 % speed-up in the neighbourhood search program part.
The agent removal is not considerably slower, however it is a very small part of the overall compute time.

The influence on the overall runtime is very small, however the memory footprint is considerably smaller.

Open for discussion on which part of the PR, we want to use!

evac_15k_timng_comp

libsimulator/src/NeighborhoodSearch.hpp
```diff
@@ -128,8 +128,37 @@ class NeighborhoodSearch
...
+
+        const auto radiusSquared = radius * radius;
+
+        for(int32_t x = xMin; x <= xMax; ++x) {
+            for(int32_t y = yMin; y <= yMax; ++y) {
+                auto it = _grid.find({x, y});
+                if(it != _grid.cend()) {
+                    for(const auto& item : it->second) {
+                        if(DistanceSquared(item->pos, pos) <= radiusSquared) {
+                            result.emplace_back(const_cast<Value*>(item));
```
Ozaq: This is unused, why should we add this code then?

libsimulator/src/GenericAgent.hpp
```diff
@@ -60,6 +61,10 @@ struct GenericAgent {
     {
     }
 };
+
+template <class Agent>
+using agent_container_t = std::deque<Agent>;
+
```
Ozaq, line 67: You can drop the template parameter since this is always used with `GenericAgent` anyways.

Please consider a different name, the current style uses PascalCase naming for types. E.g. `AgentContainer`

libsimulator/src/Simulation.cpp
```diff
@@ -338,6 +337,7 @@ void Simulation::SwitchAgentJourney(
 std::vector<GenericAgent::ID> Simulation::AgentsInRange(Point p, double distance)
 {
     JPS_SCOPED_TIMER_AND_TRACE(_timer, "Agents in Range", Debug);
+    _neighborhoodSearch.Update(_agents);
```
Ozaq: We cannot do this, `AgentsInRange` need to remain performant because we expect this to be called many times per iteration. Once intention is that users may use this to implement agent behavior changes based on location.

libsimulator/src/Simulation.cpp
```diff
@@ -358,7 +358,7 @@ std::vector<GenericAgent::ID> Simulation::AgentsInPolygon(const std::vector<Poin
         throw SimulationError("Polygon needs to be simple and convex");
     }
     const auto [p, dist] = poly.ContainingCircle();
-
+    _neighborhoodSearch.Update(_agents);
```
Ozaq: We cannot do this, `AgentsInPolygon` need to remain performant because we expect this to be called many times per iteration. Once intention is that users may use this to implement agent behavior changes based on location.

libsimulator/src/GenericAgent.hpp
pitsteinbach: I adopt the naming.

Needs to stay a templated type, as the AgentRemovalSystem is currently a templated class.

libsimulator/src/NeighborhoodSearch.hpp
pitsteinbach: this was included to run the tests in the discussion, however using the refs instead of the Agents lead to no improvement so I remove it.

Ozaq (approved):
Looks good!

Please clean up the commits as we can merge this.

PR #1597, opened 2026-06-01
Generalize routing to allow several implementations

ralfleibold:
Generalize routing implementation.

- Create a  "RoutingEngine" base class with appropriate interface as pure virtual methods.
- Move the current single implementation to "AStarRoutingEngine".
- Abstract the interface in pybind11 wrapper to use "trampoline" approach to allow implementing routing in Python.
- As a side effect, routing engine needs a `set_geometry` method. This gets called automatically by `Simulation` class. As a consequence, `RoutingEngine`s will we default-construcatble and `set_geometry` will initialize the internal structures needed for routing. The reason is that creating a `Simulation` instance should allow changing the routing engine already, but we do want to avoid initializing with the same geometry twice - and also to enforce the user to pass the geometry to both the `Simulation` class and the `RoutingEngine`. It is also strictly necessary that those are in sync.
- Extend Visualizer tool to allow changing the routing engine via dropdown menu.

TODO:

- Right now there is some `DirectRoutingEngine` show-case implementation. But this is of no use for real simulations. Move this to tests.
- Ensure routing engines have enough information available to e.g. implement TAStar-like routing in Python. E.g. do we have enough exposure from Geometry in Python?

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1597/


ralfleibold:
This work was based on the 2D interface and we move towards 3D. Therefore this is obsolete now. We might use similar ideas though in a future abstraction of the routing engine.

PR #1598, opened 2026-06-11
Python operational model

pitsteinbach:
This PR implements the necessary interface and bindings to define operational models in Python and use them in conjunction with the remaining JuPedSim C++ core.
The following components have been implemented:

- a C++ PythonModel class along with Update and Data classes
- a trampoline class that allows to define the necessary C++ functions using Python
- Python bindings for Neighbourhood search
- extend the python bindings of geometry 
- Pure Python classes for the Data and update classes, the update is performed on the C++ side, the user does not need to implement this logic in Python
- a Python wrapper class that serves as abstract base class for new user-defined models, this class wraps the native C++ binding classes to pure Python classes
- an example Python implementation of the Social Force operational model

Currently missing is the extensive documentation of this new feature. The Python docstrings have been added.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1598/


Ozaq:
Superseded by #1604

PR #1599, opened 2026-06-15
Remove switch geometry feature

Ozaq:
Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1599/


ralfleibold (dismissed):
Surprised we did not have a a single test for this. One can argue that with this, we do not need any UniqueId in Geometry object anymore. Each simulation will only hold a single geometry object over its lifetime. No blocker though - rather a general design decision when and where we store those IDs.

Ozaq:
> Surprised we did not have a a single test for this. One can argue that with this, we do not need any UniqueId in Geometry object anymore. Each simulation will only hold a single geometry object over its lifetime. No blocker though - rather a general design decision when and where we store those IDs.

good catch, will remove it

ralfleibold (approved):
Fine to be merged once GH Actions have successfully finished. Thanks!

PR #1600, opened 2026-06-18
Remove operational model cloning

Ozaq:
Move operational model ownership through the Python bindings with py::smart_holder and std::unique_ptr transfer into Simulation. This removes the unused Clone requirement from the operational model interface while keeping mesh/routing cloning unchanged.

Models are now moved out of their Python holder when creating the Simulation. The moved-from native model object is only created by the public jupedsim wrapper and is not exposed through the public Python API, so invalidating it after Simulation construction is a conscious trade-off.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1600/

### DIFF
--- a/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModel.cpp
+++ b/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModel.cpp
@@ -18,7 +18,6 @@
 #include <cstdint>
 #include <cstdlib>
 #include <limits>
-#include <memory>
 #include <numeric>
 #include <vector>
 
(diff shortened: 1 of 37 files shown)


ralfleibold (changes requested):
This works overall and looks fine. Some questions:

- AFAIK we only use `build()` of all ModelBuilder classes internally. Maybe we should return a `std::unique_ptr` directly there?
- Otherwise I think we will call the copy constructor of the model underneath when moving this via `std::make_unique`? We could probably also `std::forward` the result of `build()`, but I feel direct creation to `std::unique_pointer` would be more explicit.
- Do models still need all copy and move constructors with this change? Ideally I would like to remove at least the copy constructor.

ralfleibold (approved):
Thanks for the changes.

PR #1601, opened 2026-06-22
Move orientation from GenericAgent to ModelData

ralfleibold:
Move orientation from GenericAgent to ModelData and remove it from SoialForceModel as it is derived there from velocity. Still expose orientation in Agent as read-only property as so far all models do provide it - though not necessarily as class attributes.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1601/

### DIFF
--- a/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
+++ b/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
@@ -6,7 +6,7 @@
 #include <fmt/core.h>
 
 struct AnticipationVelocityModelData {
-    Point orientation{1.0, 0.0};
+    Point orientation{0.0, 0.0};
     double strengthNeighborRepulsion{8.0};
     double rangeNeighborRepulsion{0.1};
     double wallBufferDistance{0.1}; // buff distance of agent to wall
--- a/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
+++ b/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
@@ -6,6 +6,7 @@
 #include <fmt/core.h>
 
 struct AnticipationVelocityModelData {
+    Point orientation{0.0, 0.0};
     double strengthNeighborRepulsion{8.0};
     double rangeNeighborRepulsion{0.1};
     double wallBufferDistance{0.1}; // buff distance of agent to wall
@@ -15,6 +16,29 @@ struct AnticipationVelocityModelData {
     double timeGap{1.06};
     double v0{1.2};
     double radius{0.2};
+
+    AnticipationVelocityModelData() = default;
+    AnticipationVelocityModelData(
+        Point orientation_,
+        double strengthNeighborRepulsion_,
+        double rangeNeighborRepulsion_,
+        double wallBufferDistance_,
+        double anticipationTime_,
+        double reactionTime_,
+        double timeGap_,
+        double v0_,
+        double radius_)
+        : orientation(orientation_.Normalized())
+        , strengthNeighborRepulsion(strengthNeighborRepulsion_)
+        , rangeNeighborRepulsion(rangeNeighborRepulsion_)
+        , wallBufferDistance(wallBufferDistance_)
+        , anticipationTime(anticipationTime_)
+        , reactionTime(reactionTime_)
+        , timeGap(timeGap_)
+        , v0(v0_)
+        , radius(radius_)
+    {
+    }
 };
 
 template <>
@@ -27,9 +51,10 @@ struct fmt::formatter<AnticipationVelocityModelData> {
     {
         return fmt::format_to(
             ctx.out(),
-            "AnticipationVelocityModel[strengthNeighborRepulsion={}, "
+            "AnticipationVelocityModel[orientation={}, strengthNeighborRepulsion={}, "
             "rangeNeighborRepulsion={}, wallBufferDistance={}, "
             "timeGap={}, v0={}, radius={}, reactionTime={}, anticipationTime={}, velocity={}])",
+            m.orientation,
             m.strengthNeighborRepulsion,
             m.rangeNeighborRepulsion,
             m.wallBufferDistance,
--- a/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
+++ b/libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
@@ -6,7 +6,7 @@
 #include <fmt/core.h>
 
 struct AnticipationVelocityModelData {
-    Point orientation{0.0, 0.0};
+    Point orientation{1.0, 0.0};
     double strengthNeighborRepulsion{8.0};
     double rangeNeighborRepulsion{0.1};
     double wallBufferDistance{0.1}; // buff distance of agent to wall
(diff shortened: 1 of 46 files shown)


libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
```diff
@@ -15,6 +16,29 @@ struct AnticipationVelocityModelData {
...
+        , strengthNeighborRepulsion(strengthNeighborRepulsion_)
+        , rangeNeighborRepulsion(rangeNeighborRepulsion_)
+        , wallBufferDistance(wallBufferDistance_)
+        , anticipationTime(anticipationTime_)
+        , reactionTime(reactionTime_)
+        , timeGap(timeGap_)
+        , v0(v0_)
+        , radius(radius_)
+    {
+    }
```
Ozaq: Adding a constructor here does not add any capability to the code, neither are there any invariants that need to be protected or dependent fields who needs to be set together.

Adding constructors the structure no longer counts as an aggregate and we loose aggregate initialization, aka

```
const auto foo = AnticipationVelocityModelData{
    .orientation_ = {1.0, 0.0}
};

```

Please remove them.

libsimulator/src/OperationalModels/GeneralizedCentrifugalForceModel/GeneralizedCentrifugalForceModelData.hpp
```diff
@@ -2,9 +2,11 @@
 #pragma once
 
 #include "Point.hpp"
+#include "SimulationError.hpp"
 
 #include <fmt/core.h>
 struct GeneralizedCentrifugalForceModelData {
+    Point orientation{0.0, 0.0};
```
Ozaq, line 8: A better default is to set it to `{1.0, 0.0}`.

libsimulator/src/OperationalModels/CollisionFreeSpeedModelV3/CollisionFreeSpeedModelV3Data.hpp
```diff
@@ -4,6 +4,7 @@
 #include <fmt/core.h>
 
 struct CollisionFreeSpeedModelV3Data {
+    Point orientation{0.0, 0.0};
```
Ozaq: A better default is to set it to `{1.0, 0.0}`.

libsimulator/src/OperationalModels/CollisionFreeSpeedModelV2/CollisionFreeSpeedModelV2Data.hpp
```diff
@@ -4,6 +4,7 @@
 #include <fmt/core.h>
 
 struct CollisionFreeSpeedModelV2Data {
+    Point orientation{0.0, 0.0};
```
Ozaq, line 7: A better default is to set it to `{1.0, 0.0}`.

libsimulator/src/OperationalModels/CollisionFreeSpeedModel/CollisionFreeSpeedModelData.hpp
```diff
@@ -4,9 +4,16 @@
 #include <fmt/core.h>
 
 struct CollisionFreeSpeedModelData {
+    Point orientation{0.0, 0.0};
```
Ozaq, line 7: A better default is to set it to `{1.0, 0.0}`.

libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
```diff
@@ -6,6 +6,7 @@
 #include <fmt/core.h>
 
 struct AnticipationVelocityModelData {
+    Point orientation{0.0, 0.0};
```
Ozaq, line 9: A better default is to set it to `{1.0, 0.0}`.

libsimulator/src/Simulation.cpp
```diff
@@ -221,13 +221,6 @@ GenericAgent::ID Simulation::AddAgent(GenericAgent agent)
         throw SimulationError("Unknown stage id: {}", agent.stageId);
     }
 
-    if(std::holds_alternative<GeneralizedCentrifugalForceModelData>(agent.model))
-        if(agent.orientation.isZeroLength()) {
-            throw SimulationError(
-                "Orientation is invalid: {}. Length should be 1.", agent.orientation);
-        }
-
-    agent.orientation = agent.orientation.Normalized();
```
Ozaq, line 230: Please move the orientation is unit length check into the respective CheckModeConstraints method.

python_bindings_jupedsim/anticipation_velocity_model.cpp
```diff
@@ -33,16 +35,18 @@ void init_anticipation_velocity_model(py::module_& m)
...
                         double radius) {
                 return AnticipationVelocityModelData{
-                    .strengthNeighborRepulsion = strengthNeighborRepulsion,
-                    .rangeNeighborRepulsion = rangeNeighborRepulsion,
-                    .wallBufferDistance = wallBufferDistance,
-                    .anticipationTime = anticipationTime,
-                    .reactionTime = reactionTime,
-                    .timeGap = timeGap,
-                    .v0 = desiredSpeed,
-                    .radius = radius};
```
Ozaq, line 43: please restore the version with designated initializers

python_bindings_jupedsim/anticipation_velocity_model.cpp
```diff
@@ -51,6 +55,9 @@ void init_anticipation_velocity_model(py::module_& m)
             py::arg("time_gap"),
             py::arg("desired_speed"),
             py::arg("radius"))
+        .def_property_readonly(
+            "orientation",
+            [](const AnticipationVelocityModelData& obj) { return intoTuple(obj.orientation); })
```
Ozaq: This should also be RW

libsimulator/src/Simulation.cpp
Ozaq, line 230: Please restore the version with designated initializers for all Model Data structures.

python_bindings_jupedsim/anticipation_velocity_model.cpp
```diff
@@ -51,6 +55,9 @@ void init_anticipation_velocity_model(py::module_& m)
             py::arg("time_gap"),
             py::arg("desired_speed"),
             py::arg("radius"))
+        .def_property_readonly(
```
Ozaq: This should be RW on all models, all model state is writable

python_modules/jupedsim/jupedsim/simulation.py
```diff
@@ -299,10 +299,17 @@ def add_agent(
             Returns:
                 Id of the added agent.
         """
+
+        def orientation_or_zero(param):
+            if hasattr(param, "orientation"):
+                return param.orientation
+            return (0.0, 0.0)
+
```
Ozaq: Since orientation is now a model specific field it needs to be represented in the parameters for each model.

Ozaq (changes requested):

libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModelData.hpp
ralfleibold: Done. Using the `CheckModelConstraint` and the decision that a normalized orientation is up to the user to provide, the additional constructors are not needed anymore.

libsimulator/src/OperationalModels/GeneralizedCentrifugalForceModel/GeneralizedCentrifugalForceModelData.hpp
ralfleibold: Only changed for GCFM as it is the only one requiring a non-zero initialized orientation. All others I kept with the default that was also set in Simulation via the orientation_or_zero helper function.

libsimulator/src/Simulation.cpp
ralfleibold: Done. Thanks for the suggestion. I was not aware of this method.

python_bindings_jupedsim/anticipation_velocity_model.cpp
ralfleibold: Done.

python_bindings_jupedsim/anticipation_velocity_model.cpp
ralfleibold: Done using the new type_casters.hpp which auto-maps `Point`<->`tuple<double, double>`.

python_modules/jupedsim/jupedsim/simulation.py
ralfleibold: This is just a helper function for properly initializing the orientation for models supporting it. Added a corresponding comment. `orientation` is already part of the model parameters for each model.

Ozaq (approved):

PR #1602, opened 2026-06-22
Expose neighborhood and geometry queries to Python

Ozaq:
Add Python bindings and wrappers for querying neighboring agents from NeighborhoodSearch and nearby geometry line segments from Geometry. These APIs are needed by Python operational models to inspect their local environment during model evaluation.

Also bind LineSegment, add a Point type caster for direct tuple conversion, and add a generic binding-side range-to-vector helper for exposing iterator ranges.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1602/

### DIFF
--- a/python_bindings_jupedsim/CMakeLists.txt
+++ b/python_bindings_jupedsim/CMakeLists.txt
@@ -1,25 +1,28 @@
 add_library(py_jupedsim MODULE
+    agent.cpp
+    anticipation_velocity_model.cpp
     bindings_jupedsim.cpp
     build_info.cpp
-    conversion.cpp
-    conversion.hpp
-    generalized_centrifugal_force_model.cpp
     collision_free_speed_model.cpp
     collision_free_speed_model_v2.cpp
     collision_free_speed_model_v3.cpp
-    anticipation_velocity_model.cpp
-    social_force_model.cpp
-    warp_driver_model.cpp
+    conversion.cpp
+    conversion.hpp
+    generalized_centrifugal_force_model.cpp
+    geometry.cpp
+    journey.cpp
+    linesegment.cpp
     logging.cpp
     logging.hpp
-    trace.cpp
-    geometry.cpp
+    neighborhood_search.cpp
     routing.cpp
     simulation.cpp
-    agent.cpp
+    social_force_model.cpp
     stage.cpp
-    journey.cpp
+    trace.cpp
     transition.cpp
+    type_casters.hpp
+    warp_driver_model.cpp
 )
 
 target_link_libraries(py_jupedsim
--- a/python_bindings_jupedsim/conversion.hpp
+++ b/python_bindings_jupedsim/conversion.hpp
@@ -3,7 +3,10 @@
 
 #include <Point.hpp>
 
+#include <iterator>
+#include <ranges>
 #include <tuple>
+#include <type_traits>
 #include <vector>
 
 std::tuple<double, double> intoTuple(const Point& p);
@@ -19,10 +22,22 @@ std::vector<std::tuple<double, double>> intoTuples(const auto& in)
     return tuples;
 }
 
-
 Point intoPoint(const std::tuple<double, double>& p);
 
 std::vector<Point> intoPoints(const std::vector<std::tuple<double, double>>& in);
+
+template <typename Range>
+auto intoVec(Range&& range)
+{
+    using Value = std::remove_cvref_t<decltype(*std::begin(range))>;
+
+    std::vector<Value> result{};
+    result.reserve(std::ranges::size(range));
+    for(auto&& value : range) {
+        result.emplace_back(value);
+    }
+    return result;
+}
 
 template <typename T, typename U>
 std::vector<T> intoVecT(const std::vector<U>& vec)
--- a/python_bindings_jupedsim/geometry.cpp
+++ b/python_bindings_jupedsim/geometry.cpp
@@ -19,15 +19,23 @@ void init_geometry(py::module_& m)
             [](const CollisionGeometry& geo) {
                 return intoTuples(std::get<0>(geo.AccessibleArea()));
             })
-        .def("holes", [](const CollisionGeometry& geo) {
-            const auto holes = std::get<1>(geo.AccessibleArea());
-            std::vector<std::vector<std::tuple<double, double>>> res{};
-            res.reserve(holes.size());
-            for(const auto& hole : holes) {
-                res.emplace_back(intoTuples(hole));
-            }
-            return res;
-        });
+        .def(
+            "holes",
+            [](const CollisionGeometry& geo) {
+                const auto holes = std::get<1>(geo.AccessibleArea());
+                std::vector<std::vector<std::tuple<double, double>>> res{};
+                res.reserve(holes.size());
+                for(const auto& hole : holes) {
+                    res.emplace_back(intoTuples(hole));
+                }
+                return res;
+            })
+        .def("linesegments_close_to", &CollisionGeometry::LineSegmentsInApproxDistanceTo)
+        .def(
+            "linesegments_in_distance_to",
+            [](const CollisionGeometry& geo, double distance, std::tuple<double, double> pos) {
+                return intoVec(geo.LineSegmentsInDistanceTo(distance, intoPoint(pos)));
+            });
     py::class_<GeometryBuilder>(m, "GeometryBuilder")
         .def(py::init<>())
         .def(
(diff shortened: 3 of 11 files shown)


python_bindings_jupedsim/conversion.hpp
```diff
@@ -19,11 +21,22 @@ std::vector<std::tuple<double, double>> intoTuples(const auto& in)
...
 Point intoPoint(const std::tuple<double, double>& p);
 
 std::vector<Point> intoPoints(const std::vector<std::tuple<double, double>>& in);
 
+template <typename Range>
+auto intoVec(Range&& range)
+{
+    using Value = std::remove_cvref_t<decltype(*std::begin(range))>;
+
+    std::vector<Value> result{};
```
ralfleibold, line 34: Please also call `reserve()` with `size()` here as done in the `intoVecT` below. (I am assuming we pass in standard containers here.)

python_bindings_jupedsim/conversion.hpp
```diff
@@ -19,11 +21,22 @@ std::vector<std::tuple<double, double>> intoTuples(const auto& in)
     return tuples;
 }
 
-
 Point intoPoint(const std::tuple<double, double>& p);
 
 std::vector<Point> intoPoints(const std::vector<std::tuple<double, double>>& in);
 
+template <typename Range>
+auto intoVec(Range&& range)
```
ralfleibold, line 30: This looks nice and generic. IIRC this covers the other `intoVectT` below and can just replace it. So please remove the other function below.

python_bindings_jupedsim/neighborhood_search.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+namespace py = pybind11;
+
+void init_neighborhood_search(py::module_& m)
+{
+    py::class_<NeighborhoodSearch<GenericAgent>>(m, "NeighborhoodSearch")
+        .def(py::init<double>(), py::arg("cell_size"))
```
ralfleibold: We do not need the Python constructor as we will never construct this in Python. We cannot even add agents - a Python constructed object would be useless. JuPedSim will only get already "populated" NeighborhoodSearch instances.

python_bindings_jupedsim/geometry.cpp
```diff
@@ -19,15 +19,27 @@ void init_geometry(py::module_& m)
...
+            [](const CollisionGeometry& geo) {
+                const auto holes = std::get<1>(geo.AccessibleArea());
+                std::vector<std::vector<std::tuple<double, double>>> res{};
+                res.reserve(holes.size());
+                for(const auto& hole : holes) {
+                    res.emplace_back(intoTuples(hole));
+                }
+                return res;
+            })
+        .def(
```
ralfleibold: Using your new type_casters.hpp, you can directly map the method pointer instead of using a lambda.

```
.def("linesegments_close_to", &CollisionGeometry::LineSegmentsInApproxDistanceTo)
```

python_bindings_jupedsim/linesegment.cpp
```diff
@@ -0,0 +1,20 @@
...
+#include "type_casters.hpp" // IWYU pragma: keep
+
+#include <pybind11/pybind11.h>
+
+namespace py = pybind11;
+
+void init_linesegment(py::module_& m)
+{
+    py::class_<LineSegment>(m, "LineSegment")
+        .def(py::init<>())
```
ralfleibold: Are constructors for LineSegment and read/write exposure of p1/p2 necessary in Python? IIUC we only get line segments, but do not crate new ones for Python.
If so: Please remove constructors and make p1/p2 read-only properties.

python_modules/jupedsim/jupedsim/linesegment.py
```diff
@@ -0,0 +1,47 @@
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+import jupedsim.native as py_jps
+
+
+class LineSegment:
```
ralfleibold, line 6: We can add the Python doc to the Python bindings and remove this file as this class does nothing else than exposing the native interface as-is.

python_modules/jupedsim/jupedsim/neighborhood.py
```diff
@@ -0,0 +1,119 @@
...
+        ... )
+    """
+
+    def __init__(self, obj: py_jps.NeighborhoodSearch):
+        """
+        Wrap an existing C++
+        """
+        self._obj = obj
+
+    # def add_agent(self, agent: Agent) -> None:
```
ralfleibold: Remove commented out code.

python_modules/jupedsim/jupedsim/neighborhood.py
```diff
@@ -0,0 +1,119 @@
...
+Pure Python wrapper for neighborhood search C++ bindings.
+
+This module provides a Pythonic interface to the underlying C++ NeighborhoodSearch
+implementation, making it easier to work with spatial queries in operational models.
+"""
+
+from typing import List, Tuple
+
+import jupedsim.native as py_jps
+
```
pitsteinbach: for propper import, and running tests


```suggestion
from __future__ import annotations

from typing import TYPE_CHECKING, List, Tuple

import jupedsim.native as py_jps

if TYPE_CHECKING:
    from jupedsim.agent import Agent
```

python_modules/jupedsim/jupedsim/neighborhood.py
```diff
@@ -0,0 +1,119 @@
...
+    # Clears the old structure and rebuilds it with the given agents.
+
+    # Args:
+    # agents: List of GenericAgent objects
+
+    # Example:
+    # >>> agents = [agent1, agent2, agent3]
+    # >>> neighborhood.update(agents)
+    # """
+    # self._obj.update(agents)
```
pitsteinbach: get rid off unused member functions

```suggestion
```

pitsteinbach (commented):
I would do these two changes first

python_modules/jupedsim/jupedsim/neighborhood.py
```diff
@@ -0,0 +1,119 @@
...
+        from jupedsim.agent import Agent
+
+        if radius < 0:
+            raise ValueError(f"radius must be non-negative, got {radius}")
+
+        unwrapped_list = self._obj.get_neighboring_agents(position, radius)
+
+        native_agent_list = []
+        for cpp_agent in unwrapped_list:
+            native_agent_list.append(Agent(cpp_agent))
```
ralfleibold: List comprehensions are more readable:
```python
return [Agent(agent) for agent in unwrapped_list]
```

ralfleibold (changes requested):
Thanks. Please apply the suggested changes.

python_bindings_jupedsim/CMakeLists.txt
```diff
@@ -1,25 +1,25 @@
...
+    geometry.cpp
+    journey.cpp
+    linesegment.cpp
     logging.cpp
     logging.hpp
-    trace.cpp
-    geometry.cpp
+    neighborhood_search.cpp
     routing.cpp
     simulation.cpp
```
pitsteinbach, line 19: removed form sources
```suggestion
    simulation.cpp
    social_force_model.cpp
```

python_bindings_jupedsim/CMakeLists.txt
```diff
@@ -1,25 +1,25 @@
 add_library(py_jupedsim MODULE
     bindings_jupedsim.cpp
```
pitsteinbach, line 4: anticipating model was also removed from src list 

```suggestion
    anticipation_velocity_model.cpp
    bindings_jupedsim.cpp
```

pitsteinbach (commented):
these source files were dropped, hence the missing symbols error in tests

python_bindings_jupedsim/conversion.hpp
Ozaq: No they serve slightly different purposes, `intoVecT` converts the element type (if convertible) from `U` -> `T`, i.e. `std::vector<int8_t>` to `std::vector<uint16_t>` for example.

python_bindings_jupedsim/linesegment.cpp
Ozaq: Will be removed

python_modules/jupedsim/jupedsim/linesegment.py
Ozaq: No, we always shim user facing types because IDE integration generally refuses to auto load native extensions for security reasons. User shall be able to benefit from autocompletion and inline doc from their IDE without additional configuration.

ralfleibold (approved):
Thanks!

PR #1603, opened 2026-06-22
Custom Operational Model C++ API

pitsteinbach:
To extend JuPedSim's capabilities to add new Operational Models from application code or use our own Python bindings to implement new operational models in pure Python code. 

I added a new C++ API for the state and update classes, as they are part of a std::variant union the classes need to constructible at compile time. These classes (`CustomModelData` and `CustomModelUpdate`) are only shells that carry a std::shared_ptr with the actual implementation (`ICustomModelDataImpl`and `ICustomModelUpdateImpl`).
The user defines classes based on the Base class, for example for `ICustomModelDataImpl` the user has to implement the `void apply_update(const CustomModelUpdate& upd)` method as well as a `std::string print()` method that concatenates the model state into a string.
The pure virtual base class `ICustomModelUpdateImpl` defines the methods `std::any extract_attributes()` This method acts a function that allows to get all attributes stored in the update class and provide it to the Data class to be set in the update step.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1603/


libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
```diff
@@ -0,0 +1,47 @@
...
+    virtual std::string print() const = 0;
+    virtual void apply_update(const CustomModelUpdate& upd) = 0;
+};
+
+class CustomModelData
+{
+public:
+    explicit CustomModelData(std::shared_ptr<ICustomModelDataImpl> impl) : impl_(std::move(impl)) {}
+    ~CustomModelData() = default;
+    std::string print() const
```
ralfleibold: Rename to sth like `to_string`. `print` suggests output to console.
```suggestion
    std::string to_string() const
```

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include "CustomModelUpdate.hpp"
+
+class ICustomModelDataImpl
+{
+public:
+    virtual ~ICustomModelDataImpl() {};
+    virtual std::string print() const = 0;
```
ralfleibold:
```suggestion
    virtual std::string to_string() const = 0;
```

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
```diff
@@ -0,0 +1,47 @@
...
+    std::shared_ptr<ICustomModelDataImpl> impl_;
+};
+
+template <>
+struct fmt::formatter<CustomModelData> : fmt::formatter<std::string> {
+    template <typename FormatContext>
+    auto format(const CustomModelData& data, FormatContext& ctx) const
+    {
+        std::string repr = data.print();
+        return fmt::formatter<std::string>::format(repr, ctx);
```
ralfleibold: This is not less readable and might avoid a temporary copy of `repr` depending on what `format` allows as parameters. (Note: Suggestion already uses change to  `to_string` instead of `print`.)
```suggestion
        return fmt::formatter<std::string>::format(data.to_string(), ctx);
```

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
```diff
@@ -0,0 +1,47 @@
...
+public:
+    virtual ~ICustomModelDataImpl() {};
+    virtual std::string print() const = 0;
+    virtual void apply_update(const CustomModelUpdate& upd) = 0;
+};
+
+class CustomModelData
+{
+public:
+    explicit CustomModelData(std::shared_ptr<ICustomModelDataImpl> impl) : impl_(std::move(impl)) {}
```
ralfleibold, line 17: As we "move" in the "impl" anyways: Please change to `std::unique_ptr` if there is nothing I am missing.

libsimulator/src/OperationalModels/CustomModel/CustomModelUpdate.hpp
```diff
@@ -0,0 +1,39 @@
...
+    std::any extract_attributes() const
+    {
+        if(impl_) {
+            return impl_->extract_attributes();
+        }
+        return {};
+    }
+
+private:
+    std::shared_ptr<ICustomModelUpdateImpl> impl_;
```
ralfleibold, line 39: As in `CustomModelData`:
```suggestion
    std::unique_ptr<ICustomModelUpdateImpl> impl_;
```

ralfleibold (changes requested):
Please apply suggestions.

libsimulator/src/OperationalModels/CustomModel/CustomModelUpdate.hpp
pitsteinbach: Since copies are constructed in other parts of the code a unique_ptr can not be used here.

ralfleibold (dismissed):
Thanks.

Ozaq:
Closed in favor of #1604

PR #1604, opened 2026-06-23
Add CustomModel C++ operational model API

Ozaq:
Introduce type-erased CustomModelData and CustomModelUpdate payloads, wire them into agent/update variants, and add tests for formatting and operational update flow.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1604/


Ozaq:
@ralfleibold after discussion with @pitsteinbach this PR supersedes #1603

libsimulator/test/TestCustomModel.cpp
```diff
@@ -0,0 +1,181 @@
...
+            MinimalUpdate{position, orientation, state.velocity, state.applications + 1}};
+    }
+
+    void ApplyUpdate(const OperationalModelUpdate& update, GenericAgent& agent) const override
+    {
+        const auto& customUpdate = std::get<CustomModelUpdate>(update).Get<MinimalUpdate>();
+        auto& state = std::get<CustomModelData>(agent.model).Get<MinimalState>();
+
+        agent.pos = customUpdate.position;
+        agent.orientation = customUpdate.orientation;
```
pitsteinbach: since we are in the process of transferring orientation to the state class as well, we should include this change here already.

pitsteinbach (changes requested):
We should already move orientation out of agent and into state, other than that looks good. Also nice to have tests in C++.

libsimulator/src/OperationalModels/CustomModel/FormatAny.hpp
```diff
@@ -0,0 +1,48 @@
...
+template <typename T>
+concept HasToString = requires(const T& value) {
+    { value.ToString() } -> ToStringResult;
+};
+
+template <typename T>
+auto makeFormatFn()
+{
+    using Stored = std::decay_t<T>;
+    if constexpr(fmt::is_formattable<Stored, char>::value) {
```
ralfleibold: Checked the doc and AFAIK there is
```suggestion
    if constexpr(fmt::formattable<Stored, char>) {
```

libsimulator/src/OperationalModels/CustomModel/CustomModelUpdate.hpp
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
```
ralfleibold, line 1: If I have not missed sth, this is actually the same class definition as `CustomModelData`, just a different name. Any issue with one implementation would have to be fixed in both implementations - which can easily be forgotten. A common base class does not work. I did some search and I think one can use the concept of "phantom types / tags" like in https://github.com/insooth/insooth.github.io/blob/master/tagged-lambda-injection.md#making-phantom-types: 

```cpp
template <typename Tag>
class AnyHolder
{
private:
    std::any value{};
    FormatFn format{};

public:
    template <typename T>
        requires(!std::is_same_v<std::decay_t<T>, AnyHolder>)
    AnyHolder(T&& value) : value(std::forward<T>(value)), format(makeFormatFn<T>())
    {
        using Stored = std::decay_t<T>;
        static_assert(
            std::is_copy_constructible_v<Stored>,
            "AnyHolder payloads must be copy-constructible");
    }

    template <typename T>
    T& Get() { return std::any_cast<T&>(value); }

    template <typename T>
    const T& Get() const { return std::any_cast<const T&>(value); }

    template <typename T>
    void Set(T&& newValue)
    {
        using Stored = std::decay_t<T>;
        std::any_cast<Stored&>(value) = std::forward<T>(newValue);
    }

    friend struct fmt::formatter<AnyHolder>;
};

template <typename Tag>
struct fmt::formatter<AnyHolder<Tag>> {
    constexpr auto parse(format_parse_context& ctx) { return ctx.begin(); }

    auto format(const AnyHolder<Tag>& value, fmt::format_context& ctx) const
    {
        return value.format(value.value, ctx);
    }
};

// The two distinct types:
using CustomModelUpdate = AnyHolder<struct CustomModelUpdateTag>;
using CustomModelData   = AnyHolder<struct CustomModelDataTag>;
```

ralfleibold (changes requested):
Pretty cool stuff. The only part I want to change is the removal of code duplication. Please check my suggestion.

pitsteinbach (dismissed):

libsimulator/src/OperationalModels/CustomModel/FormatAny.hpp
```diff
@@ -0,0 +1,96 @@
...
+        };
+    } else {
+        return [](const std::any& any, fmt::format_context& ctx) {
+            const auto& value = std::any_cast<const Stored&>(any);
+            const auto addr = static_cast<const void*>(std::addressof(value));
+            return fmt::format_to(ctx.out(), "<{}@{}>", any.type().name(), addr);
+        };
+    }
+}
+template <typename Tag>
```
ralfleibold: Missing newline.

libsimulator/src/OperationalModels/CustomModel/FormatAny.hpp
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
```
ralfleibold, line 1: We should rename the file now that this is basically the core of `AnyHolder` (--> rename to AnyHolder.hpp). We could in theory keep a separate FormatAny.hpp, but as long as w do not use it elsewhere (and then it should not live in CustomModel folder), I would keep it inside here.

libsimulator/src/OperationalModels/CustomModel/CustomModelUpdate.hpp
```diff
@@ -0,0 +1,25 @@
...
+/// then apply them.
+///
+/// Like CustomModelData, the payload is stored by value in std::any. Payload types must be
+/// copy-constructible, access through Get<T>() requires the exact stored type, and mismatches throw
+/// std::bad_any_cast.
+///
+/// Formatting follows the same rules as CustomModelData: fmt formatter first, then public
+/// const-callable ToString() returning std::string, std::string_view, or const char*, then a
+/// diagnostic <type@address> fallback.
+struct CustomModelUpdateTag;
```
ralfleibold: doc probably adds all this info to `CustomModelUpdateTag`, but it should be related to `CustomModelUpdate`.

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
```diff
@@ -0,0 +1,28 @@
...
+/// value-copy semantics.
+///
+/// Access is runtime-typed: Get<T>() must use the exact stored type T. A type mismatch throws
+/// std::bad_any_cast.
+///
+/// Formatting is best-effort. If the payload has a fmt formatter, that is used. Otherwise, if it
+/// provides a public ToString() callable on const T& and returning std::string, std::string_view,
+/// or const char*, ToString() is used. If neither is available, formatting falls back to a
+/// diagnostic <type@address> representation.
+struct CustomModelDataTag;
```
ralfleibold: doc probably adds all this info to `CustomModelDataTag`, but it should be related to "CustomModelData`.

ralfleibold (changes requested):
Minor rearrangements.

@Ozaq Should we keep separate files for `CustomModelData` or `CustomeModelUpdate`? This makes it easier to find if searched for, but it is only 2 lines of code each - except for doc. And doc probably better moves to the "AnyHolder" template? As our C++ part is internal, I would remove these files and just add comments to the `#include` lines which class we are interested in.

libsimulator/test/TestCustomModel.cpp
Ozaq: done

libsimulator/src/OperationalModels/CustomModel/FormatAny.hpp
Ozaq: done

libsimulator/src/OperationalModels/CustomModel/CustomModelUpdate.hpp
Ozaq: This is a good idea but I would like to take a longer look with you and @pitsteinbach into replacing XXModelUpdate altogether and change the core loop to just generation N -> generation N+1 (There are a few consequences we need to think of but I believe this is a more promising long term route (aka road to 2.0)

libsimulator/src/OperationalModels/CustomModel/FormatAny.hpp
Ozaq: Reason to keep it here is, only place of use. Id prefer to keep it here.

libsimulator/src/OperationalModels/CustomModel/FormatAny.hpp
Ozaq: fixed

libsimulator/src/OperationalModels/CustomModel/CustomModelUpdate.hpp
Ozaq: There is not automatic doc extraction from C++

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
Ozaq: There is not automatic doc extraction from C++

libsimulator/src/OperationalModels/OperationalModelType.hpp
```diff
@@ -8,5 +8,31 @@ enum class OperationalModelType {
...
+            return "AnticipationVelocityModel";
+        case OperationalModelType::SOCIAL_FORCE:
+            return "SocialForceModel";
+        case OperationalModelType::WARP_DRIVER:
+            return "WarpDriverModel";
+        case OperationalModelType::CUSTOM_MODEL:
+            return "CustomModel";
+    }
+    return "Unknown";
+}
```
ralfleibold, line 38: If I read the code correctly, `OperationalModelType` was obsolete in the current code beforehand. (I think it was an artifact of the C-API.) Now we make it a first-class citizen again. But is there anything we cannot do the same via the `std::variant`? And for the strings: I think they are only needed for errors - and then we might rather want to enforce some `name` property/method?

python_bindings_jupedsim/python_model.cpp
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        throw SimulationError(
+            "State returned by compute_new_position() has attribute '{}' of wrong type: "
+            "expected tuple[float, float], got {} ({})",
+            attr_name,
+            actualType,
+            valueRepr);
+    }
+    customModelData.Set(pythonUpdate);
```
ralfleibold, line 157: Can we type-check that the classes on Python side(!) are the same for `customModelData` and `pythonUpdate`? IIUC we need this constraint as a new updated state object should be returned.

ralfleibold (changes requested):
I would have expected an `_apply_update()` method to be implemented on the Python side that gets the current model state. Maybe not that easy to do.

--> Current design is fine to start playing with and gain experience. 

Some changes requested: Basically if we really "reactivate" the `OperationalModelType` and whether we can increase type safety on Python side for the applied update.

libsimulator/src/OperationalModels/OperationalModelType.hpp
Ozaq: If you look at commit this was introduced you will see the OperationalModelType is used to check if Model and ModelData match. Prior to the introduction this would fail when constraint checks were done. Now the OperationalModelType is used to tie OperationalModel and OperationModelData together so that we can ensure in AddAgent to have received the correct ModelData type.

python_bindings_jupedsim/python_model.cpp
Ozaq: We could do that but as long as they are "compatible" it should be fine... I mean I cannot see a scenario where using different types per iteration is useful but that's another problem.

ralfleibold (dismissed):
Thanks for the clarifications. Approved.

pitsteinbach (dismissed):

PR #1607, opened 2026-07-03
Add 3D routing

ralfleibold:
Constraints:

- This is completely isolated from real JuPedSim engine - simulation does not use it and cannot use it.
- Take geometry from .obj file.
- Use CGAL routing: Probably does not meet our final efficiency constraints, but we do have an "oracle" engine to compare with.
- Safeguard correctness with tests.

Additional bonus in this PR:

- Viewer based on trame (https://github.com/Kitware/trame-vtk) to visualize the 3D geometry and routing.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1607/


libsimulator/src/RoutingEngine3D.hpp
```diff
@@ -0,0 +1,73 @@
...
+protected:
+    /// Result of projecting a query point onto the surface.
+    struct FaceLocation {
+        SurfaceMesh::Face_index face;
+        SurfaceKernel::Point_3 point;
+    };
+
+    /// Find face and point on face projecting via -z onto mesh.
+    /// `SurfaceMesh::null_face()` in `face` if no such point is found.
+    FaceLocation face_below(const Point3D& p) const;
```
Ozaq: This should return a location

libsimulator/src/RoutingEngine3D.hpp
Ozaq, line 1: I think this should better be split into an Interface and an implementation inheriting from this interface right from the start. Also common code should be reused through composition over inheritance right from the start changing this later is cumbersome and time consuming.

libsimulator/src/RoutingEngine3D.hpp
```diff
@@ -0,0 +1,73 @@
...
+    /// @return tuple of (path, cost): the path includes source as first element
+    ///         and the target as last element; cost is typically geodesic distance
+    ///         along it, but not necessarily (e.g. floor fields with slowness field).
+    virtual std::tuple<std::vector<Point3D>, double> get_shortest_path(const Point3D& source) = 0;
+
+    /// Get orientation to next point of the shortest path from @p source to the
+    /// current target, projected to x/y.
+    /// @param source where to route from
+    /// @return 2D orientation to the next waypoint
+    virtual Point get_orientation(const Point3D& source);
```
Ozaq: I think they need to be combined into `GetShortestPath / GetOrientation` both taking the target as input.

libsimulator/src/RoutingEngine3D.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+    // Non-copyable and non-movable
+    RoutingEngine3D(const RoutingEngine3D&) = delete;
+    RoutingEngine3D& operator=(const RoutingEngine3D&) = delete;
+    RoutingEngine3D(RoutingEngine3D&&) = delete;
+    RoutingEngine3D& operator=(RoutingEngine3D&&) = delete;
+
+    /// Set geometry.
+    /// @param mesh surface mesh to set geometry to
+    void set_geometry(SurfaceMesh&& mesh);
```
Ozaq: Why is this not done on construction?

Ozaq (commented):
Since this is targeting the integration branch I do not require you to address any of this but have a look and decide for yourself. I think the remarks are easier addressed immediately.

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: The idea is that this is a convenience function that returns both the CGAL face ID and the Point_3 in one shot. For point based algorithms we typically use the point, but for e.g. floor fields, the face ID will allow us direct access to e.g. the gradient. A `location` would require another call.
IIUC `Location` should also be a union of Point and polygon - and I do not think we ever need a "polygon" as return value.

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: I am fine creating an interface first and then deriving from it with common code.
Do you have an example of the "composition over inheritance" part?

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: This is the optimization we discussed: If we set the target each time, then any pre-computation has to be doene again for each target (like static floor field or also here for the `SurfaceMeshShortestPathRoutingEngine` in this PR the CGAL pre-computation). We can implement some cache for targets though and maybe this is the way to go given we typically deal with several waypoints - and require a new interface to invalidate the cache as otherwise it might just grow.

Do you agree with my proposed change (introducing a target cache)?

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: I think I was still thinking in terms of switching geometries - which we removed recently. I will change towards constructor argument.

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: Moved into specific `SurfaceMeshShortestPathRoutingEngine`, therefore internal now.

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: Done. `RoutingEngine3D` is now a pure interface without any implementation or member variables.

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: Done. No cache implemented right now. It was decided that this is an optimization to be done later.

libsimulator/src/RoutingEngine3D.hpp
ralfleibold: Done.

PR #1608, opened 2026-07-07
Add mesh validation functions

Ozaq:
Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1608/


libsimulator/src/Geometry/Validation.hpp
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+#pragma once
+
+#include "CfgCgal.hpp"
+
+/// @file Validation.hpp
+/// Contains support code to validate inputs or intemediate results.
```
pitsteinbach:
```suggestion
/// Contains support code to validate inputs or intermediate results.
```

libsimulator/src/Geometry/Validation.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+/// Test if the supplied normal is less tan or equal to 50deg inclined over the X/Y plane.
```
pitsteinbach:
```suggestion
/// Test if the supplied normal is less than or equal to 50deg inclined over the X/Y plane.
```

libsimulator/src/Geometry/Validation.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+template <typename MeshT>
+bool AllFacesInMeshPlanar(const MeshT& mesh)
+{
+    std::vector<VertexDescriptor<MeshT>> buffer{};
+    buffer.reserve(3);
+    return std::all_of(
+        std::begin(CGAL::faces(mesh)), std::end(CGAL::faces(mesh)), [&buffer, &mesh](auto face) {
+            buffer.clear();
+            return is_face_planar(face, mesh, buffer);
```
pitsteinbach: I guess this is a CGAL function.

Why do we need the other functions, if we can test the whole mesh with this function?

libsimulator/src/CfgCgal.hpp
```diff
@@ -18,12 +19,31 @@
...
+using Vector3 = K::Vector_3;
 
 using Poly = CGAL::Polygon_2<K>;
 using PolyWithHoles = CGAL::Polygon_with_holes_2<K>;
 using PolyWithHolesList = std::list<PolyWithHoles>;
 using PolyList = std::list<Poly>;
 using Vb = CGAL::Triangulation_vertex_base_2<K>;
+using CGALMesh = CGAL::Surface_mesh<Point3>;
+
+/// Convenienve alias for vertex descriptor types
```
pitsteinbach:
```suggestion
/// Convenience alias for vertex descriptor types
```

libsimulator/src/CfgCgal.hpp
```diff
@@ -18,12 +19,31 @@
...
+template <class G>
+using VertexDescriptor = typename boost::graph_traits<G>::vertex_descriptor;
+
+/// Convenienve alias for face descriptor types
```
pitsteinbach:
```suggestion
/// Convenience alias for face descriptor types
```

libsimulator/src/CfgCgal.hpp
```diff
@@ -18,12 +19,31 @@
...
+template <class G>
+using FaceDescriptor = typename boost::graph_traits<G>::face_descriptor;
+
+/// Convenienve alias for half edge descriptor types
```
pitsteinbach:
```suggestion
/// Convenience alias for half edge descriptor types
```

libsimulator/src/CfgCgal.hpp
```diff
@@ -18,12 +19,31 @@
...
+template <class G>
+using HalfedgeDescriptor = typename boost::graph_traits<G>::halfedge_descriptor;
+
+/// Convenienve alias for face descriptor types
```
pitsteinbach:
```suggestion
/// Convenience alias for face descriptor types
```

pitsteinbach (changes requested):
Currently, it is now clear how this will work and which functions are needed, see comment on the `AllFacesInMeshPlanar` function.

Could you add some tests to show usage, this would also help in checking that compilation works. This seems to be especially relevant since you used templating without using template instantiation in the cpp file.

libsimulator/src/Geometry/Validation.cpp
Ozaq: I expected to use this in a few other places where I am only dealing with a single polygon. But yeah I can collapse this for now and split it out when its actually used.

Ozaq:
@pitsteinbach: I removed the `is_face_planar` function and de-templated the code. I noticed that I was stilly relying on methods only implemented for SufaceMesh. Since this is conceptually the thing we want to work with just using SurfaceMesh should be fine (and easier to follow)

PR #1611, opened 2026-07-09
Cleanup

Ozaq:
Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1611/


libsimulator/src/OperationalModels/OperationalModel.hpp
```diff
@@ -82,13 +55,18 @@ class OperationalModel
     virtual ~OperationalModel() = default;
 
     virtual OperationalModelType Type() const = 0;
-    virtual OperationalModelUpdate ComputeNewPosition(
+
+    /// Computes the agent state for the next iteration.
+    /// "next" arrives as an exact copy of "current"; implementations overwrite only the fields
+    /// they change. Other agents must be read exclusively from the frozen current generation,
+    /// i.e. via "current" and the neighborhood search, never via "next".
+    virtual void ComputeNext(
```
ralfleibold: I suggest to show in the name already that we update the state.
```suggestion
    virtual void ComputeNextState(
```

libsimulator/src/OperationalModels/OperationalModel.hpp
```diff
@@ -82,13 +55,18 @@ class OperationalModel
...
         double dT,
-        const GenericAgent& ped,
+        const GenericAgent& current,
```
ralfleibold, line 65: By contract `next` contains `current" already, right? Why do we need to pass `current` at all? Is this due to Python and that we cannot ensure this is a real copy and we do not want to modify the current state implicitly?

libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModel.hpp
```diff
@@ -3,41 +3,54 @@
...
 #include <cstdint>
 #include <random>
 #include <vector>
 
-struct GenericAgent;
-
 class AnticipationVelocityModel : public OperationalModel
 {
 public:
-    using NeighborhoodSearchType = NeighborhoodSearch<GenericAgent>;
```
ralfleibold, line 20: This was no bad idea to do and it should not be removed. But the definition should have lived inside `OperationModel` previously already. Or we already remove the "templated" part of it as I do not envision us to have different types for `NeighborhoodSearch`.

libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModel.hpp
```diff
@@ -3,41 +3,54 @@
...
+    /// Per-agent state of the anticipation velocity model.
+    struct Agent {
```
ralfleibold: I suggest `State`. AFAIK this is consistent with other parts of the doc as we talk about updating states in other code locations.
```suggestion
    struct State {
```

docs/source/migration_2_0.md
```diff
@@ -0,0 +1,197 @@
...
+
+```python
+import jupedsim as jps
+
+# 1.x
+sim = jps.Simulation(model=jps.CollisionFreeSpeedModel(), geometry=area)
+
+# 2.0 — a configured model instance
+sim = jps.Simulation(model=jps.CollisionFreeSpeedModel(), geometry=area)
+```
```
ralfleibold: This is the same line in 1.x and 2.0. Please check whether this doc is still up-to-date with the latest changes.

docs/source/pedestrian_models/collision_free_speed_model_v3.md
```diff
@@ -186,9 +186,26 @@ form persistent deadlocks.
 
 ## Per-agent parameters
 
-These are exposed via
-{class}`~jupedsim.models.CollisionFreeSpeedModelV3AgentParameters` and
-can be set or modified per agent at any time.
+These are exposed as keyword-only constructor arguments of
```
ralfleibold, line 189: While I appreciate doc corrections, those should have been pushed to a separate commit as they deserve a commit of its own for tracking the history. (This applies to the other doc changes as well -except for the migration_2_0.md.)

libsimulator/src/GenericAgent.hpp
```diff
@@ -33,33 +50,41 @@ struct GenericAgent {
     Point destination{};
     Point target{};
 
-    // Agent fields common for all models
-    Point pos{};
-
     using Model = std::variant<
```
ralfleibold: For consistency this should not be the model anymore as we only access its state here.
```suggestion
    using ModelState = std::variant<
```

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
```diff
@@ -2,6 +2,7 @@
 #pragma once
```
ralfleibold: We have removed `XModelData` everywhere and moved it into a sub-class called `State`. I suggest doing the same here for consistency and remove this file.

ralfleibold (changes requested):

notebooks/corner-SocialForce.ipynb
```diff
@@ -225,16 +221,7 @@
...
-    "trajectory_file = \"corner_SocialForces.sqlite\"  # output file\n",
-    "simulation = jps.Simulation(\n",
-    "    model=jps.SocialForceModel(),\n",
-    "    geometry=area,\n",
-    "    trajectory_writer=jps.SqliteTrajectoryWriter(\n",
-    "        output_file=pathlib.Path(trajectory_file)\n",
-    "    ),\n",
-    ")"
-   ]
+   "source": "trajectory_file = \"corner_SocialForces.sqlite\"  # output file\nsimulation = jps.Simulation(\n    model=jps.SocialForceModel(),\n    geometry=area,\n    trajectory_writer=jps.SqliteTrajectoryWriter(\n        output_file=pathlib.Path(trajectory_file)\n    ),\n)"
```
ralfleibold: The changes in the notebooks do not make it more readable. Especially it is super hard to see whether there has been any change at all. If possible, please rollback the formatting part (everything in one line) so that the real changes are visible again. This is the case for all notebooks.

python_bindings_jupedsim/anticipation_velocity_model.cpp
```diff
@@ -1,72 +1,77 @@
...
         .def(
-            py::init([](std::tuple<double, double> orientation,
+            py::init([](Point position,
+                        Point orientation,
                         double strengthNeighborRepulsion,
                         double rangeNeighborRepulsion,
                         double wallBufferDistance,
                         double anticipationTime,
                         double reactionTime,
+                        Point velocity,
```
ralfleibold, line 33: This is a real change in contract to add velocity here. Is this on purpose?

python_bindings_jupedsim/generalized_centrifugal_force_model.cpp
```diff
@@ -1,54 +1,50 @@
...
-            py::arg("max_neighbor_interaction_distance"),
-            py::arg("max_geometry_interaction_distance"),
-            py::arg("max_neighbor_interpolation_distance"),
-            py::arg("max_geometry_interpolation_distance"),
-            py::arg("max_neighbor_repulsion_force"),
-            py::arg("max_geometry_repulsion_force"))
-        .def("build", &GeneralizedCentrifugalForceModelBuilder::Build);
-    py::class_<GeneralizedCentrifugalForceModelData>(m, "GeneralizedCentrifugalForceModelState")
-        .def_static("_defaults", []() { return GeneralizedCentrifugalForceModelData{}; })
+            py::arg("strength_neighbor_repulsion") = 0.3,
```
ralfleibold, line 19: You mentioned already in our call that you care about it: But this should do the same trick with "default model `d`" as in other models. It is a really nice idea - and please apply this to all the other models. (I will not add a comment for each of them.)

python_bindings_jupedsim/python_model.cpp
```diff
@@ -86,31 +85,27 @@ PythonModel::PythonModel(py::object model) : _model(std::move(model))
...
     py::gil_scoped_acquire gil;
 
-    py::object pythonAgent = py::cast(agent);
+    py::object pythonAgent = py::cast(current);
     py::object pythonGeometry = py::cast(&geometry, py::return_value_policy::reference);
     py::object pythonNeighborhoodSearch =
         py::cast((&neighborhoodSearch), py::return_value_policy::reference);
 
-    py::object update = _model.attr("_compute_new_position")(
+    py::object pythonUpdate = _model.attr("_compute_new_position")(
```
ralfleibold: This should **not** be called `_compute_new_position` anymore, but rather `_compute_new_state`.

python_bindings_jupedsim/simulation.cpp
```diff
@@ -78,7 +93,52 @@ void init_simulation(py::module_& m)
...
+            py::arg("journey_id"),
+            py::arg("stage_id"),
+            py::arg("state"))
+        .def(
+            "add_agent",
+            &AddAgentWithState<CustomModelData>,
+            py::arg("journey_id"),
+            py::arg("stage_id"),
+            py::arg("state"))
         .def(
```
ralfleibold, line 93: I have asked Claude and it came up with some definition that avoids the repetition and works automatically when we extend with more models:

```cpp
template <typename Variant>
struct AddAgentBinder;

template <typename... States>
struct AddAgentBinder<std::variant<States...>> {
    static void bind(py::class_<Simulation>& cls)
    {
        (cls.def(
             "add_agent",
             &AddAgentWithState<States>,
             py::arg("journey_id"),
             py::arg("stage_id"),
             py::arg("state")),
         ...);
    }
};
} // namespace
```
And then later you put this `AddAgentBinder`**behind** the `Simulation` pybind-class definition to add all the `add_agent` overloads:

```cpp
auto cls = py::class_<Simulation>(m, "Simulation")
    // … alle übrigen .def(...) …
    ;
AddAgentBinder<GenericAgent::ModelState>::bind(cls);
```

python_bindings_jupedsim/warp_driver_model.cpp
```diff
@@ -1,46 +1,70 @@
...
             py::arg("step_size") = 0.5,
-            py::arg("sigma") = 0.3,
             py::arg("time_uncertainty") = 0.5,
             py::arg("velocity_uncertainty_x") = 0.2,
-            py::arg("velocity_uncertainty_y") = 0.2)
-        .def("build", &WarpDriverModelBuilder::Build);
-    py::class_<WarpDriverModelData>(m, "WarpDriverModelState")
+            py::arg("velocity_uncertainty_y") = 0.2,
+            py::arg("num_samples") = 20,
+            py::arg("rng_seed") = 42);
```
ralfleibold, line 27: Once again: Contract changed compared to previous version. Intended?

python_modules/jupedsim/jupedsim/models/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+Every operational model is passed to a
+:class:`~jupedsim.simulation.Simulation` as a configured instance carrying its
+model-level parameters, e.g.
+:class:`~jupedsim.models.social_force.SocialForceModel`,
+:class:`~jupedsim.models.collision_free_speed.CollisionFreeSpeedModel` or
+:class:`~jupedsim.models.warp_driver.WarpDriverModel`. Custom Python models are
+passed as instances of a
+:class:`~jupedsim.models.custom_model.CustomOperationalModel` subclass.
+"""
```
ralfleibold, line 12: Does this add any value?

python_modules/jupedsim_examples/tests/test_python_social_force_model.py
```diff
@@ -21,12 +21,11 @@ def corridor_simulation():
...
 
 def _add_agent(sim, journey_id, stage_id, position, velocity=(0.0, 0.0)):
-    params = jps.CustomModelAgentParameters(
-        journey_id=journey_id,
-        stage_id=stage_id,
-        model=PythonSocialForceModelState(position=position, velocity=velocity),
+    return sim.add_agent(
+        journey_id,
+        stage_id,
+        PythonSocialForceModelState(position=position, velocity=velocity),
```
ralfleibold: There is no good reason IMO why to remove the clarity of keyword-arguments. Please move those back. (Same for all occurrences below.)

libsimulator/src/OperationalModels/OperationalModel.hpp
Ozaq: Will do

libsimulator/src/OperationalModels/AnticipationVelocityModel/AnticipationVelocityModel.hpp
Ozaq: Will do

docs/source/pedestrian_models/collision_free_speed_model_v3.md
Ozaq: As discussed, we will leave them in

libsimulator/src/OperationalModels/CustomModel/CustomModelData.hpp
Ozaq: Will do

libsimulator/src/OperationalModels/OperationalModel.hpp
Ozaq: I gave this a bit of thought over the weekend and I believe it makes for a nicer and clearer API to leave it at `current` and `next`. Semantically I think it is a lot clearer if the user is presented with a clear "this is your current state and it is const". Pre-initialization of `next` with `current` is IMO the actual issue but that is very handy because we have many fields who are not updated and need to be copied, no-one wants to write all that boilerplate or start every model with a `next = current`.

python_bindings_jupedsim/generalized_centrifugal_force_model.cpp
Ozaq: Those are global model parameters, we need to address them slightly differently. But this is a bit more nuanced discussion. Id like to postpone this to a followup PR. Right now we have duplicated defaults, in C++ and Python, the change you see only moves the duplication (for global model parameters) from python to the bindings. On the actual state defaults the issue is fixed.

python_bindings_jupedsim/anticipation_velocity_model.cpp
Ozaq: Yes, by design we wanted to allow setting the full state of an agent. Velocity was imply missing.

python_bindings_jupedsim/python_model.cpp
Ozaq: good catch

python_bindings_jupedsim/simulation.cpp
Ozaq: I addressed it differently but the repetition was stupid to begin with. You can simply accept a variant and move it.

python_bindings_jupedsim/warp_driver_model.cpp
Ozaq: We let users set all fields on initialization, those fields were missing.

python_modules/jupedsim/jupedsim/models/__init__.py
Ozaq: It's an explanation which becomes visible in the documentation. I think it does add some value, yes.

libsimulator/src/OperationalModels/OperationalModel.hpp
ralfleibold: Ok. Let's go with the "current"/"next" approach.

notebooks/corner-SocialForce.ipynb
```diff
@@ -291,9 +291,9 @@
     "\n",
     "for pos, v0 in zip(pos_in_spawning_area, v_distribution):\n",
     "    simulation.add_agent(\n",
-    "        jps.SocialForceModelAgentParameters(\n",
-    "            journey_id=journey_id, stage_id=exit_id, position=pos\n",
-    "        )\n",
+    "        journey_id=journey_id,\n",
+    "        stage_id=exit_id,\n",
+    "        state=jps.SocialForceModelState(position=pos, desired_speed=v0),\n",
```
ralfleibold, line 296: We have not used `v0` beforehand. Maybe this was a bug though. If so, it might be better to fix it with `desired_speed` though as our doc rate `v0` as deprecated IIRC.

notebooks/corner.ipynb
```diff
@@ -294,12 +294,9 @@
...
     "    simulation.add_agent(\n",
-    "        jps.CollisionFreeSpeedModelAgentParameters(\n",
-    "            journey_id=journey_id,\n",
-    "            stage_id=exit_id,\n",
-    "            position=pos,\n",
-    "            desired_speed=v0,\n",
-    "        )\n",
+    "        journey_id=journey_id,\n",
+    "        stage_id=exit_id,\n",
+    "        state=jps.CollisionFreeSpeedModelState(position=pos, desired_speed=v0),\n",
```
ralfleibold, line 299: Same: `v0` --> `desired_speed`

notebooks/model-comparison.ipynb
```diff
@@ -371,8 +338,8 @@
     "model = \"AVM\"\n",
     "trajectory_file = f\"{model}.sqlite\"\n",
     "simulation = initialize_simulation(\n",
-    "    jps.AnticipationVelocityModel(),\n",
-    "    jps.AnticipationVelocityModelAgentParameters,\n",
+    "    jps.AnticipationVelocityModel(rng_seed=42),\n",
```
ralfleibold, line 341: I do not see a seed beforehand. Why do we do now?

ralfleibold (approved):
I would like to see these minor changes, but this is not blocking. Up to you whether you still apply or merge as-is.

notebooks/model-comparison.ipynb
Ozaq: AVM did not set a seed value.

notebooks/corner-SocialForce.ipynb
Ozaq: `v0` is deprecated in the codebase in an attempt to make the model code easier to read. But v0 is still the term used in the original papers. The Examples usually use v0. This code is just moved.

notebooks/model-comparison.ipynb
Ozaq: To make the examples that can be configured with a seed match. (E.g. warp driver )

ralfleibold: Ok, good change to make this explicit.

PR #1615, opened 2026-07-14
Rework operational models around flat per-agent state

pitsteinbach:
Split the per-agent model state structs out of the model classes into GenericAgentState.hpp and move agent position and journey data (AgentJourney) into that state. Operational models now implement GetNeighbors() plus ComputeNextState() on the GenericState variant and receive the frozen neighbor states instead of the whole agent / neighborhood search. Locals prefer 'state' over 'model' naming throughout the model implementations.

PythonModel keeps the Python-facing callback signature by capturing the neighborhood search handle in GetNeighbors and rebuilding a transient agent view for the callback.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1615/


libsimulator/src/GenericAgentState.hpp
Ozaq, line 1: I understand why you needed to remove the agent state from the concrete Operational Models. Int his case I want you to place the concrete model states each into their own header in the respective subfolder instead of this one.

libsimulator/src/OperationalDecisionSystem.hpp
```diff
@@ -38,8 +38,18 @@ class OperationalDecisionSystem
     {
         _next.clear();
         std::copy(std::begin(agents), std::end(agents), std::back_inserter(_next));
+        OperationalModel::StateContainer neighbor_states;
         for(size_t index = 0; index < agents.size(); ++index) {
-            _model->ComputeNextState(dT, agents[index], _next[index], geometry, neighborhoodSearch);
+            neighbor_states.clear();
+            _model->GetNeighbors(
+                agents[index].state, neighborhoodSearch, geometry, neighbor_states);
```
Ozaq: This is not model specific and there should be only one implementation. The implementation in the models currently (on master) differs due to code drift / omissions. 

This needs to be one implementation either as free function or on the OperationalModelDecisionSystem. This way you can undo the pushdown of id into the model state. ID does semantically not belong into the ModelState because the ID identifies the whole Entity (aka Generic Agent -> which we probably could rename to just Agent now)

libsimulator/src/GenericAgentState.hpp
```diff
@@ -0,0 +1,357 @@
...
+// The variant that represents the per-agent operational model state.
+using GenericAgentModel = std::variant<
+    GcfmState,
+    CfsmState,
+    CfsmV2State,
+    CfsmV3State,
+    AvmState,
+    SfmState,
+    WdmState,
+    CustomModelState>;
```
Ozaq: This should move to its own header under OperationalModels, e.g. OperationalModelState. Then the name should follow the new pattern of <Model>State, i.e. OperationalModelState. This would tie in nicely with TacticalModelState / StrategicModelState (see other comment)

libsimulator/src/TacticalModelState.hpp
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+#pragma once
+
+#include "Point.hpp"
+
+struct AgentJourney {
+    Point target;
+    Point destination{};
+};
```
Ozaq, line 8: This is now a combination of two separate things: The state of the strategic and the state of the tactical model.

I propose to replace this with:

```
struct StrategicModelState {
    JourneyId journeyId; // strictly user controlled
    StageId stageId; // controlled by StrategicModel 
    Point target; // controlled by StrategicModel 
};

struct TacticalModelState{
    Point target // formerly destination, controlled by TacticalModel
};

So (Generic)Agent becomes:

class Agent {
    StrategicModelState sms; // think of a better name
    TacticalModelState tms;
    OperationalModelState oms;
}

```

Ozaq (changes requested):

PR #1616, opened 2026-07-14
Add Geometry3D including RegionSplit.

ralfleibold:
- Add `Geometry3D` class that holds a CGAL `SufaceMesh` and a lot of utility methods.
- Support RegionSplit: Split `Geometry3D` into regions where each x/y position is unique in each region. This means `(x, y)` + `region_id` can be transformed into `(x, y, z)` and vice versa.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1616/


libsimulator/src/Geometry/RegionSplit.cpp
```diff
@@ -0,0 +1,132 @@
...
+        return poly->size() >= 3; // convex polygon overlap
+    }
+    return false; // Point_2 / Segment_2: boundary touch only
+}
+} // namespace
+
+RegionSplit split_into_regions(SurfaceMesh& mesh)
+{
+    auto [region, added] =
+        mesh.add_property_map<SurfaceMesh::Face_index, std::size_t>("f:region", 0);
```
Ozaq: Use: 

```cpp
auto region = mesh.add_property_map<SurfaceMesh::Face_index, std::size_t>("f:region", 0).first;
```

libsimulator/src/Geometry/Geometry3D.hpp
```diff
@@ -0,0 +1,105 @@
...
+    };
+
+    /// Take an already-built surface mesh (e.g. from a mesh builder or a test).
+    explicit Geometry3D(SurfaceMesh&& mesh);
+
+    /// Build from a 2D walkable area, lifted flat to z=0. Uses the same
+    /// constrained Delaunay triangulation as the 2D RoutingEngine, so the 2D
+    /// and 3D pipelines run on the identical triangle set -- the basis for
+    /// exact parity comparisons. Also keeps the 2D view (collision_geometry()).
+    explicit Geometry3D(const PolyWithHoles& poly);
```
Ozaq: Just take it by value unless there is any real indication that the additional move ctor has any performance impact

Ozaq (approved):

libsimulator/src/Geometry/Geometry3D.hpp
ralfleibold: Good suggestion. Done.

libsimulator/src/Geometry/RegionSplit.cpp
ralfleibold: Done. As discussed, I have fixed the initialization part (`add_property_map` with default value `UNASSIGNED`) while changing this line.

Ozaq (approved):

PR #1618, opened 2026-07-15
Move cgal 2d kernel to EPICK

ralfleibold:
We use a different kernel for current 2D JuPedSim. In the multi-floor scenario with 3D coordinates, we need a different kernel ("EPICK") due to the SurfaceMeshShortestPath routing. this PR harmonize this to "all EPICK".
In addition, change `Point2` to `Point2D` to make this consistent with `Point3D` and to express the dimension clearly in the type name itself.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1618/


ralfleibold:
Dropped this PR as this should not merge to "master", but to the multi-floor-building-support branch

PR #1626, opened 2026-07-20
Add the next waypoint to the agent python binding

### LINKED ISSUE #1625: Python custom-model layer cannot access `agent.destination` (the routed waypoint)
## Summary

A Python operational model written against `CustomOperationalModel` receives
`agent.target` (the journey/stage goal) but has no access to `agent.destination`
(the routed next waypoint). Every built-in C++ operational model uses
`destination`; the Python layer is the only model in the tree that cannot see it.

Checked against `master` at 1cb8b6fda.

## Detail

The two fields are set in different systems during `Simulation::Iterate`:

`StrategicalDecisionSystem` sets the stage goal:

```cpp
const auto [target, id] = journeys.at(agent.journeyId)->Target(agent);
agent.target = target;
```

`TacticalDecisionSystem` then computes the routed waypoint from it:

```cpp
const auto dest = agent.target;
agent.destination = routingEngine.ComputeWaypoint(agent.position(), dest);
```

So `destination` is what an operational model is meant to steer toward, and
indeed all seven built-in models read it -- `AnticipationVelocityModel`,
`CollisionFreeSpeedModel` (V1/V2/V3), `GeneralizedCentrifugalForceModel`,
`SocialForceModel`, `WarpDriverModel`. None of them read `agent.target`.

But the Python binding in `python_bindings_jupedsim/agent.cpp` exposes only:

```
id, journey_id, stage_id, position, target, model
```

`destination` is not bound, so a Python model can only see the final goal.

## Impact

A custom Python model cannot navigate non-convex geometry. Pointing agents at
`target` sends them at the final goal in a straight line, through walls.

Minimal reproduction using the `corner.ipynb` geometry -- an agent in the lower
corridor is told to head at the exit around the bend:

```python
import jupedsim as jps
from shapely import Polygon

AREA = Polygon([(0, 0), (12, 0), (12, 12), (10, 12), (10, 2), (0, 2)])
EXIT = Polygon([(10, 11), (12, 11), (12, 12), (10, 12)])

class Probe(jps.models.custom_model.CustomOperationalModel):
    def compute_next_state(self, dt, ped, geometry, neighborhood_search):
        print(dir(ped))          # no 'destination'
        print(ped.position, ped.target)
        return ped.model
```

At position `(2.0, 1.0)` this prints `target = (11.0, 11.5)` -- the exit centroid,
constant, straight through the corner wall. With no access to the waypoint there
is no way to recover the route from inside the model.

## Related question

Is exposing the *remaining route* (rather than only the next waypoint) something
you would consider? The routing engine already computes the path internally. Some
learned models use a path-length-vs-straight-line ratio as an input feature, which
needs the full path rather than just the next waypoint. Entirely understand if the
intended contract is "agents only ever know their next waypoint" -- knowing that
either way would settle the design on our side.

pitsteinbach:
Fixes #1625.

The changes include (exceeding the scope of #1625):
- Expose not only the final target, but also the next target of the agent to the Python bindings.
- Rename `target` to `final_target` and `destination` to `next_target` to clarify the meaning. This adds to the readability/clarity of the code.
- Fix the Python SocialForce model implementation as it used `final_target` instead of (the previously non-existing) `next_target`.

It should be noted that exposing `next_target` is only temporary. It is likely that we will change API to have an `orientation` towards the next target, instead of the location itself.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1626/

### DIFF
--- a/docs/source/concepts/routing.rst
+++ b/docs/source/concepts/routing.rst
@@ -337,7 +337,7 @@ It will move towards the specified point and if the point is reached, it will co
 .. code:: python
 
     agent = simulation.agent(direct_steering_agent_id)
-    agent.end_destination = (-10, -10)
+    agent.final_target = (-10, -10)
 
 .. note::
 
--- a/docs/source/concepts/routing.rst
+++ b/docs/source/concepts/routing.rst
@@ -337,7 +337,7 @@ It will move towards the specified point and if the point is reached, it will co
 .. code:: python
 
     agent = simulation.agent(direct_steering_agent_id)
-    agent.target = (-10, -10)
+    agent.end_destination = (-10, -10)
 
 .. note::
 
--- a/python_modules/jupedsim/jupedsim/agent.py
+++ b/python_modules/jupedsim/jupedsim/agent.py
@@ -133,8 +133,8 @@ class Agent:
         return self.__resolve().position
 
     @property
-    def end_destination(self) -> tuple[float, float]:
-        """Current end destination of the agent.
+    def end_target(self) -> tuple[float, float]:
+        """Current end target of the agent.
 
         Can be used to directly steer an agent towards the given coordinate.
         This will bypass the strategical and tactical level, but the
@@ -152,17 +152,17 @@ class Agent:
             iteration call.
 
         Returns:
-            Current end destination of the agent.
+            Current end target of the agent.
         """
-        return self.__resolve().end_destination
+        return self.__resolve().end_target
 
-    @end_destination.setter
-    def end_destination(self, end_destination: tuple[float, float]) -> None:
-        self.__resolve().end_destination = end_destination
+    @end_target.setter
+    def end_target(self, end_target: tuple[float, float]) -> None:
+        self.__resolve().end_target = end_target
 
     @property
-    def next_destination(self) -> tuple[float, float]:
-        """Current next destination of the agent.
+    def next_target(self) -> tuple[float, float]:
+        """Current next target of the agent.
 
         Next destination is the next waypoint of the current stage of the agent's journey.
         It is used by the operational model to compute the next state of the agent.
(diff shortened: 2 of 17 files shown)


docs/source/concepts/routing.rst
```diff
@@ -337,7 +337,7 @@ It will move towards the specified point and if the point is reached, it will co
 .. code:: python
 
     agent = simulation.agent(direct_steering_agent_id)
-    agent.target = (-10, -10)
+    agent.end_destination = (-10, -10)
```
pitsteinbach:
```suggestion
    agent.final_target = (-10, -10)
```

python_modules/jupedsim/jupedsim/agent.py
```diff
@@ -133,8 +133,8 @@ def position(self) -> tuple[float, float]:
         return self.__resolve().position
 
     @property
-    def target(self) -> tuple[float, float]:
-        """Current target of the agent.
+    def final_target(self) -> tuple[float, float]:
+        """Current end target of the agent.
```
pitsteinbach:
```suggestion
        """Current final target of the agent.
```

python_modules/jupedsim/jupedsim/agent.py
```diff
@@ -152,13 +152,25 @@ def target(self) -> tuple[float, float]:
             iteration call.
 
         Returns:
-            Current target of the agent.
+            Current end target of the agent.
```
pitsteinbach:
```suggestion
            Current final target of the agent.
```

python_modules/jupedsim/jupedsim/agent.py
```diff
@@ -216,9 +228,14 @@ def position(self) -> tuple[float, float]:
         return self.__obj.position
 
     @property
-    def target(self) -> tuple[float, float]:
-        """Current target of the agent."""
-        return self.__obj.target
+    def final_target(self) -> tuple[float, float]:
+        """Current end target of the agent."""
```
pitsteinbach:
```suggestion
        """Current final target of the agent."""
```

ralfleibold (approved):
Nice changes, @pitsteinbach . Thanks!

FabianPlum:
Sweet! Thanks for the swift updates here!
Will continue my integration tomorrow!

PR #1627, opened 2026-07-21
New interface for neighbour and geometry queries

pitsteinbach:
This pull request introduces a new `EnvironmentQuery` utility class to centralize and simplify agent/geometry queries in the simulation. The main operational models (`AnticipationVelocityModel`, `CollisionFreeSpeedModel`, and `CollisionFreeSpeedModelV2`) are refactored to use this new class, leading to clearer, more maintainable code. The changes also update the operational decision system to pass `EnvironmentQuery` instead of separate geometry and neighborhood search objects.

**Core refactor: EnvironmentQuery abstraction**

- Introduced the `EnvironmentQuery` class (`libsimulator/src/EnvironmentQuery.hpp`) to encapsulate common environment queries, such as finding agents in range, checking visibility, and querying geometry.
- Updated all operational models (`AnticipationVelocityModel`, `CollisionFreeSpeedModel`, `CollisionFreeSpeedModelV2`) to use `EnvironmentQuery` for neighborhood and visibility checks, replacing manual filtering and geometry intersection logic. [[1]](diffhunk://#diff-adfddcb5ff6fbbae78f61873e0afcaeb17fd5533e32ede14eaa7f46b1a602f02L37-R42) [[2]](diffhunk://#diff-7508e8309e2bb2a21127c30f2249500c64fbf3158f9cabea96aa51584dcbff71L42-R46) [[3]](diffhunk://#diff-1571fad2f256bba4d4c836e349917ea82e6853b0031307c1c79a0a5cb9605ef7L30-R34)
- Changed the interface of operational models and the operational decision system to accept `EnvironmentQuery` instead of separate `CollisionGeometry` and `NeighborhoodSearch` parameters. [[1]](diffhunk://#diff-c4a650d68d33b3e2e933c21527beb995d8c57c928f5a410e25fdeaf60d69136bL49-R53) [[2]](diffhunk://#diff-5e656a7e6f9ed2a1595adb3989737660e249a9bae923b498a67fe359adc354edL43-R47) [[3]](diffhunk://#diff-9cd15ae108a7fe2109a631396a2922b2161ad8f89664d637f0c9bf236f63eb45L32-R42)
- Updated model constraint checks in all models to use `EnvironmentQuery` for neighbor and geometry queries, ensuring consistent logic and reducing code duplication. [[1]](diffhunk://#diff-adfddcb5ff6fbbae78f61873e0afcaeb17fd5533e32ede14eaa7f46b1a602f02L133-R108) [[2]](diffhunk://#diff-adfddcb5ff6fbbae78f61873e0afcaeb17fd5533e32ede14eaa7f46b1a602f02L178-L182) [[3]](diffhunk://#diff-7508e8309e2bb2a21127c30f2249500c64fbf3158f9cabea96aa51584dcbff71L112-R86) [[4]](diffhunk://#diff-7508e8309e2bb2a21127c30f2249500c64fbf3158f9cabea96aa51584dcbff71L132-L136) [[5]](diffhunk://#diff-1571fad2f256bba4d4c836e349917ea82e6853b0031307c1c79a0a5cb9605ef7L100-R74)
- Added a new test file `TestEnvironmentQuery.cpp` to the build to cover the new utility class.

These changes significantly improve code clarity, reduce duplication, and make it easier to extend or modify environment queries in the future.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1627/


libsimulator/src/EnvironmentQuery.hpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    struct AcceptAll {
+        bool operator()(const GenericAgent&) const { return true; }
+    };
+
+    // Returns all agents within 'radius' of 'agent', excluding 'agent' itself.
+    // An optional predicate 'include' further filters the result; it receives each
+    // candidate agent and must return true to keep it. Example:
+    //   query.AgentsInRange(self, r, query.VisibleFrom(self.position()))
```
ralfleibold: Please change from Python example to C++ example.

libsimulator/src/EnvironmentQuery.hpp
```diff
@@ -0,0 +1,79 @@
...
+    template <std::predicate<const GenericAgent&> Pred = AcceptAll>
+    std::vector<GenericAgent>
+    AgentsInRange(const GenericAgent& agent, double radius, Pred include = {}) const
```
ralfleibold: In the sense of std lib and std::ranges I would rather like to define a "filter" than an "include" predicate. You also use it here like this as you do `erase_if`.

libsimulator/src/Simulation.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include "AgentRemovalSystem.hpp"
 #include "CollisionGeometry.hpp"
+#include "EnvironmentQuery.hpp"
```
ralfleibold: This looks obsolete as there is no other code change in the file.

libsimulator/test/TestEnvironmentQuery.cpp
```diff
@@ -0,0 +1,150 @@
...
+// Build NeighborhoodSearch populated with all agents in the container.
+// Returns both so the caller can keep both alive.
+struct Fixture {
+    AgentContainer<GenericAgent> agents{};
+    NeighborhoodSearch<GenericAgent> nsearch{5.0};
+
+    void add(GenericAgent a) { agents.push_back(std::move(a)); }
+
+    void commit() { nsearch.Update(agents); }
+};
```
ralfleibold: The names look highly non-intuitive to me. if you pas the ge, you can request the EnvironmentQuery directly. It also seems that we can pass the position only. Below is a suggestion, but feel free to change.
```suggestion
struct Environment {
    AgentContainer<GenericAgent> agents{};
    NeighborhoodSearch<GenericAgent> neightbodhood_search{5.0};
    Geometry geo;

    void add_agent(Point pos, double radius = 0.2) { agents.push_back(MakeAgent(pos, radius)); }

    EnvironmentQuery get_query_object() { neighborhood_search.update(); return {geo, neighborhood_search}; }
};
```

libsimulator/test/TestEnvironmentQuery.cpp
```diff
@@ -0,0 +1,150 @@
...
+    f.add(MakeAgent({1, 0}, 0.2)); // small radius — filtered out
+    f.add(MakeAgent({0, 1}, 0.4)); // large radius — kept
+    f.add(MakeAgent({-1, 0}, 0.4)); // large radius — kept
+    f.commit();
+
+    const auto geo = OpenGeometry();
+    const EnvironmentQuery q{geo, f.nsearch};
+
+    const auto result = q.AgentsInRange(f.agents[0], 5.0, [](const GenericAgent& a) {
+        return std::get<State>(a.model).radius > 0.25;
```
ralfleibold: For a test: Better set a `gt_radius` as constexpr and then use this instead of using 0.25 two times. This ensure consistency even if the radius is changed later on.

libsimulator/test/TestEnvironmentQuery.cpp
```diff
@@ -0,0 +1,150 @@
...
+    f.commit();
+
+    const auto geo = OpenGeometry();
+    const EnvironmentQuery q{geo, f.nsearch};
+
+    const auto selfId = f.agents[0].id;
+    bool selfSeen = false;
+    q.AgentsInRange(f.agents[0], 5.0, [&](const GenericAgent& a) {
+        if(a.id == selfId) {
+            selfSeen = true;
```
ralfleibold: Just fail here. This is more intuitive.

performancetest/evac_shortest_path.py
```diff
@@ -61,7 +61,6 @@ def main(
     number_of_iterations: int = 100,
     number_of_agents: int = 15000,
 ):
```
ralfleibold, line 63: Is this triggered by formatting? Looks completely unrelated to you PR...

python_bindings_jupedsim/environment_query.cpp
```diff
@@ -0,0 +1,82 @@
...
+               double radius,
+               py::function predicate) -> std::vector<GenericAgent> {
+                return self.AgentsInRange(
+                    agent, radius, [&predicate](const GenericAgent& candidate) -> bool {
+                        return predicate(py::cast(candidate)).cast<bool>();
+                    });
+            },
+            py::arg("agent"),
+            py::arg("radius"),
+            py::arg("predicate"),
```
ralfleibold: Does that still work with the default argument to not filter at all? If you always have to pass a filter in Python, this is not good.

python_bindings_jupedsim/bindings_jupedsim.cpp
```diff
@@ -25,6 +25,7 @@ void init_stage(py::module_& m);
 void init_simulation(py::module_& m);
 void init_neighborhood_search(py::module_& m);
```
ralfleibold: We can remove exposure of `NeighborHoodSearch` completely, right? Please do so.

python_modules/jupedsim/jupedsim/agent.py
```diff
@@ -250,5 +250,9 @@ def model(self) -> Any:
             return state.model
         return state
 
+    @property
```
ralfleibold, line 253: Why do we need this? Seems you can also just use __obj which other code does as well to pass the "native" object, doesn't it? Otherwise `_native` seems what we use in terms of wording elsewhere IIRC.

python_modules/jupedsim/jupedsim/environment_query.py
```diff
@@ -0,0 +1,150 @@
...
+                multiple predicates with ``lambda n: p1(n) and p2(n)``.
+
+        Returns:
+            List of transient agent views. Only valid during the current callback.
+        """
+        from jupedsim.agent import _TransientAgent
+
+        raw_agent = agent._raw if isinstance(agent, _TransientAgent) else agent
+
+        if predicate is None:
```
ralfleibold: As stated n the Python bindings: Can we avoid having to code this logic here?

libsimulator/src/EnvironmentQuery.hpp
```diff
@@ -0,0 +1,79 @@
...
+            return VisibilityBetween(from, candidate.position());
+        };
+    }
+
+    bool VisibilityBetween(const Point& from, const Point& to) const
+    {
+        return !_geometry.IntersectsAny(LineSegment{from, to});
+    }
+
+    const std::vector<LineSegment>& LineSegmentsInRange(const Point& p) const
```
ralfleibold: Using the same name is not good. I would expect some default radius which is not the case. We should also call it `LineSegmentsInApproxDistanceTo`. Otherwise this is confusing. Maybe we should rather rebrand it `LineSegmentsInGridCellDistance`?

Please change accordingly for the exposed Python interface.

python_bindings_jupedsim/environment_query.cpp
```diff
@@ -0,0 +1,82 @@
...
+                const Point from = intoPoint(pos);
+                return py::cpp_function([geo, from](const GenericAgent& candidate) -> bool {
+                    return !geo->IntersectsAny(LineSegment{from, candidate.position()});
+                });
+            },
+            py::arg("position"),
+            "Returns a predicate: True when position has unobstructed line-of-sight to the "
+            "candidate.")
+        .def(
+            "line_segments_in_range",
```
ralfleibold: See comment in the C++ part: It should have a different name.

python_modules/jupedsim/jupedsim/environment_query.py
```diff
@@ -0,0 +1,150 @@
...
+                _TransientAgent(a)
+                for a in self._obj.agents_in_range(raw_agent, radius)
+            ]
+
+        def _wrapped(raw):
+            return predicate(_TransientAgent(raw))
+
+        return [
+            _TransientAgent(a)
+            for a in self._obj.agents_in_range(raw_agent, radius, _wrapped)
```
ralfleibold: If we do a list comprehension, it is not even beneficial to have the filter added to the C++ call. You can simply write 
```python
return [
    _TransientAgent(a)
    for a in self._obj.agents_in_range(raw_agent, radius) if not predicate(a)
]
```
Or am I missing sth?

python_modules/jupedsim/jupedsim/environment_query.py
```diff
@@ -0,0 +1,150 @@
...
+
+            visible = env_query.visible_from(ped.position)
+            neighbors = env_query.agents_in_range(ped, 5.0, visible)
+        """
+        from jupedsim.agent import _TransientAgent
+
+        raw_pred = self._obj.visible_from(position)
+
+        def _pred(agent: _TransientAgent) -> bool:
+            raw = agent._raw if isinstance(agent, _TransientAgent) else agent
```
ralfleibold: The pattern for duck-typing and what is used in other parts of the code is rather
```suggestion
            raw = agent._raw if hasattr(agent, "_raw") else agent
```

python_modules/jupedsim/tests/test_environment_query.py
```diff
@@ -0,0 +1,257 @@
...
+def _make_sim(model, geometry=None):
+    if geometry is None:
+        geometry = _open_room()
+    sim = jps.Simulation(model=model, geometry=geometry, dt=0.05)
+    exit_id = sim.add_exit_stage([(19, 9), (19, 11), (20, 11), (20, 9)])
+    journey_id = sim.add_journey(jps.JourneyDescription([exit_id]))
+    return sim, exit_id, journey_id
+
+
+def _add(sim, journey_id, stage_id, position, group=0):
```
ralfleibold: I do not feel that this helper function does make things anyhow shorter, does it? I suggest removing it.

python_modules/jupedsim/tests/test_environment_query.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def _is_probe(self, ped):
+        px, py = ped.position
+        ppx, ppy = self._probe_pos
+        return abs(px - ppx) < 1e-4 and abs(py - ppy) < 1e-4
+
+    def compute_next_state(self, dt, ped, env_query):
+        if self._is_probe(ped):
+            if self._predicate is None:
+                neighbors = env_query.agents_in_range(ped, self._radius)
```
ralfleibold: I would simplify to track in any case and just define
```python
def _tracking(neighbor):
    self.predicate_positions.append(tuple(neighbor.position))
    return self._predicate(neighbor) if self._predicate else True
```

python_modules/jupedsim/tests/test_environment_query.py
```diff
@@ -0,0 +1,257 @@
...
+            else:
+                called = self.predicate_positions
```
ralfleibold: Don't we need to clear `self.predicate_positions`?

ralfleibold (changes requested):
This will be a great cleanup. Especially the model updates in C++ read a lot easier now. 

As discussed in person:
- Let's go with the `filter` approach in `AgenstInRange` - but restricted to the C++ API. In Python it is easier and more efficient to directly filter inside the list comprehension.
- Let's ensure we have the performance optimization of the `LineSegmentsInApproxDistanceTo` in `VisibleBetween`.

python_bindings_jupedsim/environment_query.cpp
pitsteinbach: above this function definition is the same function but without the filter argument, which triggers the default filter in C++ of returning everything. The first Python tests does exactly that `test_no_predicate_returns_all_in_radius()`.

python_modules/jupedsim/tests/test_environment_query.py
pitsteinbach: There are 21 calls to this function in the tests, which with our formatting standards saves us 4 lines of code for every call, I would say that is a good reason to use it.

python_modules/jupedsim/tests/test_environment_query.py
pitsteinbach: Every test instantiates new agent instances and then only performs a single iteration, if we would do multiple iterations then yes. I can add It so we don't forget it if we add future tests to this file?

python_bindings_jupedsim/environment_query.cpp
ralfleibold: This is refactored and this comment obsolete now.

performancetest/evac_shortest_path.py
pitsteinbach: should be yes.

python_modules/jupedsim/jupedsim/agent.py
pitsteinbach: LLM says due to the mangling of Python `__obj` is not seen as an attribute of the class, I just changed to `_native`

libsimulator/src/EnvironmentQuery.hpp
```diff
@@ -0,0 +1,87 @@
...
+    // position for which neighbors are returned as well as the candidates. Example:
+    //   query.AgentsInRange(state, r, [&envQuery, from=model.position](const Point& to) {
+    //   return envQuery.NoGeometryBetween(from, to);})
+    template <std::predicate<const Point&> Pred = AcceptAll>
+    std::vector<GenericAgent>
+    OtherAgentsInRange(const OperationalModelState& state, double radius, Pred filter = {}) const
+    {
+        Point from = Pos(state);
+        return AgentsInRange(from, radius, [&from, &filter](const Point& to) {
+            return !(from == to || !filter(to));
```
ralfleibold:
```suggestion
            return from != to && filter(to);
```

ralfleibold (dismissed):
Just a very minor change to be done. Please do a squash merge.

PR #1634, opened 2026-07-31
Modernize and improve cmake config

ralfleibold:
- Use ccache if installed. This significantly speeds up subsequent build due to the huge perfetto.cc file.
- Remove some obsolete includes, especially one time perfetto.h
- Check whether compiler supports IPO (for C++) and use a global setting rather than per-target ones.
- Move common cmake compile options into `INTERFACE` instead of a variable.
- Use common compile options for py_jupedsim as well. This revealed an issue with an unused variable --> fixed.
- Enable JuPedSim to be used as sub-project by using `PROJECT_SOURCE_DIR` / `PROJECT_BINARY_DIR`.
- Check against `WIN32` instead of comparing `CMAKE_SYSTEM` against "Windows".

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1634/


pitsteinbach:
Cool changes that clean the build system, thanks!

pitsteinbach (approved):

PR #1639, opened 2026-08-05
Avoid recomputing walls for every neighbor in range call

pitsteinbach:
In #1627, I introduced the new filter to filter out agents occluded by walls.
However, the walls where looked up every neighbor leading to a big performance penalty.

In this PR I introduce a caching behavior, that only recomputes the walls if the from location is changing.
This restored the previous computational performance.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1639/


ralfleibold (approved):
Thanks!

PR #1640, opened 2026-08-06
Make CustomModel::State constructor explicit.

ralfleibold:
- This prevents implicit conversion to CustomModel::State from any type.
- Fix corresponding issues in CheckModelConstraint of the operational models passing "position" instead of "state" which greedily converts to CustomModel::State. This caused models to check for neighbors around the origin instead of the real agent's position.
- Add corresponding tests to prevent this from happening again.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1640/


pitsteinbach (approved):
Thanks for catching that one, and nice tests to avoid this in the future.

PR #1641, opened 2026-08-10
Default random seed for AVM

pitsteinbach:
Small PR, about adding a default value for the AVM random seed (again). This must have been dropped during refactoring all other models with random generators also have default values for the seed, this looks like a refactoring "bug" to me.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1641/


ralfleibold (approved):
Thanks.

PR #1642, opened 2026-08-10
Introduce state mapping between current state and neighbor states

pitsteinbach:
This pull request introduces a mechanism for Python models to delegate simulation steps to built-in C++ models while controlling how neighbor agent states are presented. This enables seamless model composition, where a Python model can "wrap" a built-in model and ensure neighbors are seen with the expected state type, solving a key challenge for multi-model simulations. The changes add C++ and Python APIs for neighbor state mapping, and some tests.

The most important changes are:

**Core C++ API and Implementation:**

* Introduced the `NeighborStates` interface and added support to `AgentView` and `AgentStep` for substituting how neighbor states are seen, enabling custom mapping of neighbor states for delegation scenarios. [[1]](diffhunk://#diff-41e72be8cecf5b79ccd2503bed775f033467bd4153637a3147a24a363c2ce424R40-R62) [[2]](diffhunk://#diff-41e72be8cecf5b79ccd2503bed775f033467bd4153637a3147a24a363c2ce424L64-R92) [[3]](diffhunk://#diff-41e72be8cecf5b79ccd2503bed775f033467bd4153637a3147a24a363c2ce424R144-R165)
* Implemented `PythonNeighborStates`, a C++ class that wraps a Python callable to perform neighbor state mapping, and exposed it to Python via bindings.

**Python Bindings and API:**

* Added the `AgentStep.with_neighbor_state_mapping()` method to the Python API, enabling Python models to create a view where neighbors are seen through a user-provided mapping function. [[1]](diffhunk://#diff-90aeaf927335584cc7ee6be6c13c3db8b50ccee3000ba2022599f8a2c49a919dR205-R232) [[2]](diffhunk://#diff-e9bef486962cef30987502b612b80a198898a5c9d1e3886afcec399bffda3d58R108-R115)
* Updated the C++/Python API for native models to accept either a wrapped or native `AgentStep` and to provide clear error messages when state types are mismatched or mappings are incorrect.

**Testing and Validation:**

* Added a test suite (`test_model_delegation.py`) that verifies delegation, neighbor state mapping, error handling, and behavioral equivalence between delegated and native models.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1642/

### DIFF
--- a/libsimulator/src/AgentView.hpp
+++ b/libsimulator/src/AgentView.hpp
@@ -37,14 +37,29 @@ inline bool intersects(const LineSegment& los, const WallView& wall)
 {
     return intersects(los, wall.segment);
 }
+/// Substitutes the model state neighbors are seen with.
+///
+/// A model that delegates a step to another model hands over a view of the world, but the
+/// delegate expects neighbors to carry its own state type. An implementation maps each
+/// neighbor's stored state to one the delegate understands and owns the mapped states for as
+/// long as it lives.
+class NeighborStates
+{
+public:
+    virtual ~NeighborStates() = default;
+    virtual const OperationalModelState& MapToCurrentState(const GenericAgent& agent) const = 0;
+};
 
 /// What an agent perceives of its surroundings, expressed relative to where it
 /// stands. Agents do not know absolute positions as they do not need to.
 class AgentView
 {
 public:
-    AgentView(const EnvironmentQuery& world, const GenericAgent& agent)
-        : _world(world), _agent(agent)
+    AgentView(
+        const EnvironmentQuery& world,
+        const GenericAgent& agent,
+        const NeighborStates* neighborStates = nullptr)
+        : _world(world), _agent(agent), _neighborStates(neighborStates)
     {
     }
 
@@ -61,13 +76,19 @@ public:
             if(candidate.id == _agent.id) {
                 return;
             }
-            const NeighborView neighbor{candidate.Position() - _agent.Position(), &candidate.state};
+            const NeighborView neighbor{
+                candidate.Position() - _agent.Position(),
+                _neighborStates ? &_neighborStates->MapToCurrentState(candidate) :
+                                  &candidate.state};
             if(filter(neighbor)) {
                 neighbors.push_back(neighbor);
             }
         });
         return neighbors;
     }
+
+    /// Whether neighbors are seen through a substituted state rather than their own.
+    bool HasNeighborMapping() const { return _neighborStates != nullptr; }
 
     /// Whether the straight line to a point at 'RelativePosition' is free of geometry.
     /// 'boundaries' must be in the same relative coordinate frame (agent at origin),
@@ -120,15 +141,28 @@ public:
 protected:
     const EnvironmentQuery& _world;
     const GenericAgent& _agent;
+    /// Null in the common case, where neighbors are seen with their own state.
+    const NeighborStates* _neighborStates;
 };
 
 /// An AgentView plus what only holds for one step (dT + next target).
 class AgentStep : public AgentView
 {
 public:
-    AgentStep(const EnvironmentQuery& world, const GenericAgent& agent, double dt)
-        : AgentView(world, agent), _dt(dt)
+    AgentStep(
+        const EnvironmentQuery& world,
+        const GenericAgent& agent,
+        double dt,
+        const NeighborStates* neighborStates = nullptr)
+        : AgentView(world, agent, neighborStates), _dt(dt)
     {
+    }
+
+    /// The same step, but with neighbors seen through 'states'. 'states' has to outlive the
+    /// returned step.
+    AgentStep WithNeighborStateMapping(const NeighborStates& states) const
+    {
+        return AgentStep{_world, _agent, _dt, &states};
     }
 
     double dt() const { return _dt; }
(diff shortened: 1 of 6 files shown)


python_modules/jupedsim/tests/test_model_delegation.py
```diff
@@ -0,0 +1,268 @@
...
+        journey_id=journey_id,
+        stage_id=stage_id,
+        position=(1.0, 10.0),
+        state=state,
+    )
+    trajectory = [sim.agent(agent_id).position]
+    for _ in range(steps):
+        sim.iterate()
+        trajectory.append(sim.agent(agent_id).position)
+    return trajectory
```
Ozaq, line 65: This should be pytest fixtures

python_modules/jupedsim/tests/test_model_delegation.py
Ozaq, line 1: No one usually bothers with prefixing classes / methods in test code with `_` to indicate visibility. This is usually strictly done only in modules imported by "others"

python_bindings_jupedsim/agent_view.cpp
```diff
@@ -9,23 +9,61 @@
...
+#include <utility>
 #include <variant>
 
 namespace py = pybind11;
 
+namespace
+{
+py::object stateToPython(const OperationalModelState& state)
+{
+    const auto& variant = static_cast<const OperationalModelState&>(state);
```
Ozaq, line 22: The cast is a no-op

python_bindings_jupedsim/python_model.cpp
```diff
@@ -163,9 +164,58 @@ void PythonModel::CheckModelConstraint(const GenericAgent& agent, const AgentVie
...
+                            "{} encountered a neighbor it cannot read. The mapping passed to "
+                            "with_neighbor_states() has to return '{}' states." :
+                            "{} encountered a neighbor it cannot read. Map neighbors to '{}' "
+                            "states with AgentStep.with_neighbor_states() before delegating.",
+                        ToString(self.Type()),
+                        ToString(self.Type()));
+                }
+
+                return std::make_tuple(
+                    static_cast<const OperationalModelState&>(next),
```
Ozaq, line 212: the cast is a no-op

python_bindings_jupedsim/python_model.cpp
```diff
@@ -163,9 +164,58 @@ void PythonModel::CheckModelConstraint(const GenericAgent& agent, const AgentVie
...
+                try {
+                    movement = self.ComputeNextState(current, next, agentStep);
+                } catch(const std::bad_variant_access&) {
+                    // The state handed in is of the right type, so it was a neighbor's.
+                    throw SimulationError(
+                        agentStep.HasNeighborMapping() ?
+                            "{} encountered a neighbor it cannot read. The mapping passed to "
+                            "with_neighbor_states() has to return '{}' states." :
+                            "{} encountered a neighbor it cannot read. Map neighbors to '{}' "
+                            "states with AgentStep.with_neighbor_states() before delegating.",
```
Ozaq, line 206: The error text should use `WithNeighborStates` because it referees to the C++ type.

python_bindings_jupedsim/python_model.cpp
```diff
@@ -163,9 +164,58 @@ void PythonModel::CheckModelConstraint(const GenericAgent& agent, const AgentVie
     _model.attr("_check_model_constraint")(pythonState, pythonView);
 }
 
+/// The Python surface wraps the native view in jupedsim.agent_view.AgentStep. A model that
+/// delegates hands on the step it was given, so accept either the wrapper or the native object.
+static const AgentStep& asAgentStep(const py::object& step)
+{
+    const py::object native = py::hasattr(step, "_obj") ? step.attr("_obj") : step;
```
Ozaq, line 171: Better use `py::getattr(step, "_obj", step)`.

Add missing check for 'None', e.g.

```cpp
if(native.is_none()) {
    // ...error...
}
```

python_bindings_jupedsim/python_model.cpp
pitsteinbach: Yes but the conversion is given to a Python function with the corresponding name `with_neighbor_state_mapping`, I would rename it to that the current name is from the initial design. I don't think a message with a C++ function name ist helpful to the user as they never touch C++ code or functions.

libsimulator/src/AgentView.hpp
```diff
@@ -37,14 +37,29 @@ inline bool intersects(const LineSegment& los, const WallView& wall)
 {
     return intersects(los, wall.segment);
 }
+/// Substitutes the model state neighbors are seen with.
+///
+/// A model that delegates a step to another model hands over a view of the world, but the
+/// delegate expects neighbors to carry its own state type. An implementation maps each
+/// neighbor's stored state to one the delegate understands and owns the mapped states for as
+/// long as it lives.
+class NeighborStates
```
Ozaq: The name implies something static not a on demand conversion operation to me.
E.e. 'StateMapper' although more generic would make the code easier to follow.

Alternatively scrap the whole type and use a `std::function` instead.

libsimulator/src/AgentView.hpp
```diff
@@ -37,14 +37,29 @@ inline bool intersects(const LineSegment& los, const WallView& wall)
...
+{
+public:
+    virtual ~NeighborStates() = default;
+    virtual const OperationalModelState& MapToCurrentState(const GenericAgent& agent) const = 0;
```
Ozaq: Right now this method maps `Agent` and its internal `State` to another `State`. This would be clearer if it would map `State` -> `State` directly.

Ozaq (approved):

PR #1643, opened 2026-08-10
Adapt performance test 15k to new API

pitsteinbach:
The performance test `evac_shortest_path.py` was not updated during the last few API changes and was hence based on an old API. I ported the test to the current API. 
Other performance tests might need an update as well.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1643/


ralfleibold (approved):
Looks fine. Thanks.

Please squash the commits into one and adjust the commit message. "New API" is a bit too general. Maybe state that position, journey_id and stage_id and model-state are now passed separately.

PR #1652, opened 2026-09-06
Fetch citation information with zenodo-bibtex-exporter

schroedtert:
Replace the hand rolled Zenodo lookup in docs/source/_scripts with the zenodo-bibtex-exporter tool, which resolves the record by concept id instead of searching for the project name and filtering the results.

The search based lookup read a single page of results, and Zenodo caps that at 25 records for anonymous requests. JuPedSim has more published versions than that, so the older ones were already unreachable: building the documentation for v0.9.0 reported that no citation information was available, for a record that exists. Looking the version up by concept id filters server side and finds it.

The function now always returns a string. Previously it returned None whenever anything went wrong, and conf.py passes the result straight to file.write(), so a brief Zenodo outage failed the documentation build with a TypeError rather than degrading. A development version, which matches no release, took the same path on every build.

Failures now fall back to the most recent release, and an unreachable Zenodo produces a BibTeX comment pointing at the concept DOI, so the generated .bib file stays valid either way.

The signature is unchanged, so conf.py needs no adjustment.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1652/

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -16,3 +16,7 @@ plotly-express
 numpy
 pandas
 pedpy
+
+# Citation information is fetched from Zenodo at build time.
+# Pinned to a branch until the tool is released to PyPI.
+zenodo-bibtex-exporter @ git+https://github.com/schroedtert/zenodo-bibtex-exporter.git@create-zenodo-bibtex-exporter
--- a/docs/source/_scripts/generate_bibtex.py
+++ b/docs/source/_scripts/generate_bibtex.py
@@ -1,68 +1,60 @@
-import textwrap
-import time
-import warnings
+"""Fetch the citation information for this build from Zenodo.
 
-import requests
+The record is looked up by *concept id*, the identifier that stays the same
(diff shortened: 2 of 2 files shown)


docs/requirements.txt
```diff
@@ -16,3 +16,7 @@ plotly-express
 numpy
 pandas
 pedpy
+
+# Citation information is fetched from Zenodo at build time.
+# Pinned to a branch until the tool is released to PyPI.
+zenodo-bibtex-exporter @ git+https://github.com/schroedtert/zenodo-bibtex-exporter.git@create-zenodo-bibtex-exporter
```
Ozaq, line 22: @schroedtert Can we get that on PyPI first?

schroedtert: This will be done, I just need to find the time to finalize the tool and release it. Was a first POC if everything works in the repos using this tool as it should. 

If you want to take a look at that too: https://github.com/schroedtert/zenodo-bibtex-exporter/pull/1

PR #1656, opened 2026-09-17
New API to define geometries

ralfleibold:
- Introduce `WalkableSurface` class.
  - Use `add_region` to define a flat region (aka "floor").
  - Use `connect_regions` to connect floors.
- Each added region and connector is a region of its own in the resulting geometry.
- Add tests to cover error cases.
- Convert `examples/example_3d.py` to using this new API.
- `WalkableSurface` can create a `RegionGraph2D` for future usage in floor fields.

Preview: https://PedestrianDynamics.github.io/jupedsim/pull-requests/1656/


python_bindings_jupedsim/type_casters.hpp
```diff
@@ -16,15 +19,21 @@ struct type_caster<Point> {
...
         auto seq = reinterpret_borrow<sequence>(src);
         if(seq.size() != 2) {
             return false;
         }
-        value.x = seq[0].cast<double>();
-        value.y = seq[1].cast<double>();
+        auto x_caster = make_caster<double>();
+        auto y_caster = make_caster<double>();
+        // call below with "true": allow implicit type conversion to target double type
+        if(!x_caster.load(seq[0], true) || !y_caster.load(seq[1], true)) {
```
Ozaq, line 32: `load(...)` should be called with the boolean parameter of this function as 2nd argument

python_modules/jupedsim/tests/test_new_geometry.py
```diff
@@ -0,0 +1,454 @@
...
+            from_region=self.ground_floor,
+            from_edge=self.ground_conn_edge,
+            to_region=self.upper_floor,
+            to_edge=self.upper_conn_edge,
+        )
+
+
+@pytest.fixture
+def walkable_surface():
+    return jps.WalkableSurface()
```
Ozaq: That fixture is a bit pointless? Just use the constructor

python_bindings_jupedsim/type_casters.hpp
ralfleibold: pybind does not call with "convert=true" if
- if the function/method is overloaded
- AND: if there is not positional argument (only kwargs)

This is the case here and therefore it would not convert these arguments. Latest pybind11 3.1.0 fixed the issue that it converts at least int-->double now. So it is not that significant anymore. We can probably also remove this again with latest pybind11 version. I can verify this again.

Ozaq (approved):

PR #1658, opened 2026-09-21
Fixup release publish

Ozaq:
Now use new deploy-documentation workflow and removed old workflows from cibuildwheel.

New preview: https://jupedsim-pr-previews.pedestriandynamics.org/pull-requests/1658/ (built from 7dd7a01)


.github/workflows/deploy-documentation.yml
```diff
@@ -6,212 +6,124 @@ on:
...
+          for attempt in 1 2 3 4 5; do
+            git fetch origin gh-pages
+            git reset --hard origin/gh-pages
+
+            rm -rf "$TAG"
+            mkdir "$TAG"
+            cp -r "$RUNNER_TEMP/html/." "$TAG/"
+
+            rm -rf stable
+            ln -s "$TAG" stable
```
ralfleibold: Seems we always tag the latest release as the "stable" one? E.g. if we release a 1.x patch after 2.0  we mark the wrong as stable, right? Same with "-dev" releases (regex `const pattern = /^v[0-9]+\.[0-9]+\.[0-9]+(-dev\.[0-9]+)?$/;` above). "stable" should also be the one with highest major number that is not a dev-release.

Better move this logic into the "write_doc_versions.py" script? We check all versions there anywhere.

.github/workflows/deploy-documentation.yml
```diff
@@ -6,212 +6,124 @@ on:
...
-            const url = `https://${repo.owner}.github.io/${repo.repo}/${process.env.PREVIEW_DIR}/`;
-
-            const start = '<!-- PREVIEW-URL-START -->';
-            const end = '<!-- PREVIEW-URL-END -->';
-            const snippet = `${start}\nPreview: ${url}\n${end}`;
+            if (matching.length !== 1) {
+              core.setFailed(`Expected exactly one release tag at ${run.head_sha}, found: [${matching.join(', ')}]`);
+              return;
+            }
+            const tag = matching[0];
```
ralfleibold, line 46: IIUC we cannot label a "-dev" release as stable with this logic as is would point to the exact same code? If so we might want to think about a more robust solution - though I do not expect this to happen often.

.github/workflows/deploy-documentation.yml
```diff
@@ -6,212 +6,124 @@ on:
...
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git add versions.json stable
-          git commit -m "docs: update versions and stable symlink"
-          git push
+
+          # Other jobs may push to gh-pages concurrently, so a push can be
+          # rejected as non-fast-forward. On rejection, restart from the
+          # current remote tip and re-stage the release.
```
ralfleibold, line 104: Still needed? This PR changes to run only for releases - not PRs anymore. It should never run concurrently anymore.

.github/workflows/cibuildwheel.yml
ralfleibold:
```suggestion
  contents: read
```
I don't think we need writer permissions anymore after removing the "Documentation" part

ralfleibold (changes requested):

.github/workflows/deploy-documentation.yml
Ozaq: Good idea, will do

.github/workflows/deploy-documentation.yml
Ozaq: We never will tag a -dev release as stable, wtf are we? Savages? ;)

.github/workflows/deploy-documentation.yml
Ozaq: Yes this is still very much needed! if we do multiple releases in short order, e.g. a hot fix several releases they will run concurrent and try to push the new doc. They can safely retry because each targets a different path and computes the version.json based on HEAD it sees in the repo. GH concurrent control does not help us here because its either abort and use newest or queue jobs with queue length of 1. both would lead to lost deployments in specific (albeit rare) cases.

ralfleibold (dismissed):
Look good now. Thanks!

